PR #5, opened 2019-07-21
term: review usage of "source"

jorgeorpinel:
To pair up with recent changes in the [docs repo](https://github.com/iterative/dvc.org).

Also, I may need this fork in the near future to update the get started section of the docs...

### DIFF
--- a/create.sh
+++ b/create.sh
@@ -40,7 +40,7 @@ mkdir data
 wget https://dvc.org/s3/get-started/data.xml -O data/data.xml
 dvc add data/data.xml
 git add data/.gitignore data/data.xml.dvc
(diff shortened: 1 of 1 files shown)


shcheklein:
Thanks for updating this!

> Also, I may need this fork in the near future to update the get started section of the docs...

What do you mean?

jorgeorpinel:
Np!

> What do you mean?

We want to update the get started guide after https://github.com/iterative/dvc.org/pull/490 right?

shcheklein:
👍 yep, when we add a section we might want to update. It will depend on how we will update the Get started itself.

PR #6, opened 2019-08-17
finish restructuring and get-started generation updates

jorgeorpinel:
Matches https://github.com/iterative/dvc.org/pull/561

> Per https://github.com/iterative/dvc.org/issues/487#issuecomment-521320785


jorgeorpinel:
I just realized the new example-get-started/generate.sh script doesn't fully run (at least on my Mac). Let me review before this gets approved, but feel free to review the work so far @shcheklein. Thanks

jorgeorpinel:
This is ready and tested but I don't have access to delete or rewrite https://github.com/iterative/example-get-started so I can't deploy the newly generated repo with my few updates.

README.md
```diff
@@ -9,24 +9,32 @@ Started](https://dvc.org/doc/get-started) and other sections of the DVC docs.
 Please make sure you have these available on the environment where these scripts
 will run:
 
-- Git
-- Python (with `pip`)
+- [Git](https://git-scm.com/)
+- [Python](https://www.python.org/) 3 (with `python3` and [pip](https://pypi.org/project/pip/) commands)
```
shcheklein: it's actually should be possible to run it with Python 2, no? as least get started itself is compatible with Python 2

jorgeorpinel: Probably. The [script](https://github.com/iterative/example-repos-dev/pull/6/files#diff-95ce536a3300fc445a5e3964c5e6e8bcR32) already has `virtualenv -p python3 .env` though so I just documented that as a requirement. Should I change it? Or maybe just check that it works on Python 2 and not specify a version in the script and README?

jorgeorpinel:
Note I just rewrote https://github.com/iterative/example-get-started to match the changes in this PR.

example-get-started/generate.sh
```diff
@@ -5,16 +5,16 @@
 # x Print commands and their arguments as they are executed.
 set -eux
 
-THIS="$( cd "$(dirname "$0")" ; pwd -P )"
+HERE="$( cd "$(dirname "$0")" ; pwd -P )"
 REPO_NAME="example-get-started"
-REPO_PATH="../$REPO_NAME"
+REPO_PATH="$HERE/build/$REPO_NAME"
```
shcheklein: add build to gitignore?

README.md
shcheklein: no, you are right. let's keep it. It's just I'm running it manually from time to time with Python 2 to make sure that everything works as expected. But when we generate it, it's probably better to use the same version every time. I'll think if we need an option to run it w Python 2.

shcheklein:
@jorgeorpinel  I've merged, there were one-two unresolved comments - please take a look if something of them are not resolved yet.

README.md
jorgeorpinel: Lmk

example-get-started/generate.sh
jorgeorpinel: Good thinking. `build/` is already in the root .gitignore but that was a coincidence. Will add the the directory .gitignore too.

jorgeorpinel: Added in 1a9ce81 (will show up in my next PR).

PR #7, opened 2019-08-21
Add data registry repo generator [WIP]

jorgeorpinel:
Related to https://github.com/iterative/dvc.org/issues/487

### DIFF
--- a/example-get-started/.gitignore
+++ b/example-get-started/.gitignore
@@ -1,3 +1,4 @@
 # Custom
 *.zip
 /tmp
+build/


jorgeorpinel:
@shcheklein the current versions of https://github.com/iterative/dataset-registry and https://github.com/iterative/example-get-started were generated with the code in this PR.

dataset-registry/generate.sh
```diff
@@ -0,0 +1,80 @@
...
+cp $HERE/code/README.md $REPO_PATH
+
+git add .
+git commit -m "Init & config DVC project, add README"
+
+# Get Started
+
+mkdir get-started
+wget https://data.dvc.org/get-started/data.xml -O get-started/data.xml
+dvc add get-started/data.xml
```
jorgeorpinel: Question @shcheklein: Shouldn't data.xml be a zip file on our S3 bucket and dataset-registry repo remote? Like in https://dvc.org/doc/get-started/example-versioning

shcheklein: that was a conscious decision to use the plain `data.xml` in get started. The reason - simplicity (to avoid introducing an extra step with unzip). It was beneficial to include it in the tutorial to show that DVC is language agnostic, basically to show that it can run any command, not only python. 

So, I would keep it XML for the get started. We can decide later when we start consolidating tutorial and NLP example if we want to keep ZIP.

shcheklein:
> @shcheklein the current versions of https://github.com/iterative/dataset-registry and https://github.com/iterative/example-get-started were generated with the code in this PR.

👍

PR #14, opened 2020-05-10
update get-started to 1.0

shcheklein:
@jorgeorpinel a branch to play with 1.0 - it generates a project ready to use

### DIFF
--- a/example-get-started/generate.sh
+++ b/example-get-started/generate.sh
@@ -19,96 +19,73 @@ fi
 mkdir -p $REPO_PATH
 pushd $REPO_PATH
 
-# (https://dvc.org/doc/install)
-
 virtualenv -p python3 .venv
 export VIRTUAL_ENV_DISABLE_PROMPT=true
 source .venv/bin/activate
 echo '.venv/' > .gitignore
 
-
 pip install gitpython
 pip install "git+https://github.com/iterative/dvc#egg=dvc[all]"
 
-# https://dvc.org/doc/tutorials/get-started#initialize
-
 git init
+cp $HERE/code/README.md .
 git add .
 git commit -m  "Initialize Git repository"
-
 git tag -a "0-git-init" -m "Git initialized."
+
 
 dvc init
 git commit -m "Initialize DVC project"
-
 git tag -a "1-dvc-init" -m "DVC initialized."
 
-# https://dvc.org/doc/tutorials/get-started/data-versioning
 
 mkdir data
 dvc get https://github.com/iterative/dataset-registry \
         get-started/data.xml -o data/data.xml
-
-# https://dvc.org/doc/tutorials/get-started/data-versioning#start-tracking-data
-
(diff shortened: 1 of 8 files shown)


jorgeorpinel:
Thank you 🙏 this should be pretty useful. I have my own branch also... With lots of udpates per the Get Started new structure, but still using DVC 0.9x so I'll have to merge them...

example-get-started/generate.sh
```diff
@@ -26,7 +26,9 @@ export VIRTUAL_ENV_DISABLE_PROMPT=true
 source .venv/bin/activate
 echo '.venv/' > .gitignore
 
-pip install "dvc[s3]"
+
+pip install gitpython
+pip install "git+https://github.com/iterative/dvc#egg=dvc[all]"
```
jorgeorpinel, line 31: This is temporary though, right? Should just be `pip install dvc` once 1.0 is fully released.

shcheklein: it's a good question :) I would say it's even more useful for a us to keep it to the latest master?

jorgeorpinel: OK yeah maybe here in the gen, but not in the actual Get Started 🙂 

Also, only `git+https://github.com/iterative/dvc#egg=dvc[s3]` (S3) is needed here, I think.

shcheklein: oh, yeah. Not in the actual get started, that's for sure. Even here - I don't have a strong opinion to be honest. It feels it can be useful since we'll need it to test new iterations down the road.

example-get-started/generate.sh
```diff
@@ -19,146 +19,118 @@ fi
...
-
 git tag -a "2-track-data" -m "Data file added."
 
-# https://dvc.org/doc/tutorials/get-started/data-versioning#configure-remote-storage
 
-# Remote active on this env only, for writing to HTTP redirect below.
 dvc remote add -d --local storage s3://dvc-public/remote/get-started
-
-# Actual remote for generated project (read-only). Redirect of S3 bucket above.
 dvc remote add -d storage https://remote.dvc.org/get-started
```
jorgeorpinel, line 52: You didn't find these comments useful?

shcheklein: I think I removed it along other comments (that were not very useful since we have tags + hard to maintain up to date since they had links in them). Don't have a strong opinion about this one.

shcheklein:
@jorgeorpinel 

> AUC or some other basic value like that.

it has AUC, files was renamed and type changes to `scores.json`

example-get-started/generate.sh
jorgeorpinel: I liked them. Helped me understand why there were 2 remotes in this project. I presonaly don't need then any more but in general I think they're useful comments.

shcheklein: Yep, np! I'll get these two back.

PR #16, opened 2020-10-04
Update example-get-started readme

metalglove:
The README.md assumes that the user which is just getting started with dvc, already has dvc installed.
By adding this note the user will know that it can install dvc if it is missing.

### DIFF
--- a/example-get-started/code/README.md
+++ b/example-get-started/code/README.md
@@ -33,10 +33,9 @@ $ source .env/bin/activate
 $ pip install -r src/requirements.txt
 ```
 
-> This instruction assumes that DVC is already installed, as it is frequently 
-used as a global tool like Git. If DVC is not installed, see the 
-[DVC installation guide](https://dvc.org/doc/install) on how to install DVC.
-
+> This instruction assumes that DVC is already installed, as it is frequently
+> used as a global tool like Git. If DVC is not installed, see the
+> [DVC installation guide](https://dvc.org/doc/install) on how to install DVC.
 
 This DVC project comes with a preconfigured DVC
 [remote storage](https://dvc.org/doc/commands-reference/remote) that holds raw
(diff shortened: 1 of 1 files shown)


example-get-started/code/README.md
```diff
@@ -33,10 +33,17 @@ $ source .env/bin/activate
 $ pip install -r src/requirements.txt
 ```
 
+Note that dvc is not installed by the requirements file, 
+as it is supposed to be used as a global tool like git. If dvc is not installed,
+use the following command to install it:
+```console
```
shcheklein, line 39: it's better to point to the install page, let users decide how they want DVC to be installed

example-get-started/code/README.md
```diff
@@ -33,10 +33,17 @@ $ source .env/bin/activate
 $ pip install -r src/requirements.txt
 ```
 
+Note that dvc is not installed by the requirements file,
```
shcheklein, line 36: dvc -> DVC

example-get-started/code/README.md
```diff
@@ -33,10 +33,17 @@ $ source .env/bin/activate
...
+use the following command to install it:
+```console
+$ pip install dvc
+```
+
 This DVC project comes with a preconfigured DVC
 [remote storage](https://dvc.org/doc/commands-reference/remote) that holds raw
 data (input), intermediate, and final results that are produced. This is a
-read-only HTTP remote.
+read-only HTTP remote.
```
shcheklein, line 46: remove extra space

example-get-started/code/README.md
shcheklein, line 36: let's make it a 

> Note ....

and move to the very top of the section

also, text should be something along those lines:

"This instruction assumes that DVC is already installed, see the installation guide and pick the best way to install it that works for you"

shcheklein (approved):

PR #31, opened 2021-04-13
Move desc to import

dberenbaum:
Sorry, looks like the `dvc add` gets overwritten later by a `dvc import` later, so moved `--desc` to the later command to keep it in the final example repo.

### DIFF
--- a/example-get-started/generate.sh
+++ b/example-get-started/generate.sh
@@ -42,7 +42,7 @@ git tag -a "1-dvc-init" -m "DVC initialized."
 mkdir data
 dvc get https://github.com/iterative/dataset-registry \
         get-started/data.xml -o data/data.xml
-dvc add data/data.xml
+dvc add data/data.xml --desc "Initial XML StackOverflow dataset (raw data)"
 git add data/.gitignore data/data.xml.dvc
 git commit -m "Add raw data"
 git tag -a "2-track-data" -m "Data file added."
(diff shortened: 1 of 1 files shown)


example-get-started/generate.sh
```diff
@@ -19,13 +19,13 @@ fi
 mkdir -p $REPO_PATH
 pushd $REPO_PATH
 
-virtualenv -p python3 .venv
-export VIRTUAL_ENV_DISABLE_PROMPT=true
-source .venv/bin/activate
-echo '.venv/' > .gitignore
+#virtualenv -p python3 .venv
```
shcheklein, line 22: it's not intentional probably?

dberenbaum: Good catch! Removed.

example-get-started/generate.sh
```diff
@@ -42,7 +42,7 @@ git tag -a "1-dvc-init" -m "DVC initialized."
 mkdir data
 dvc get https://github.com/iterative/dataset-registry \
         get-started/data.xml -o data/data.xml
-dvc add data/data.xml --desc "Initial XML StackOverflow dataset (raw data)"
```
shcheklein, line 45: May we can actually keep both? Change the message a bit? I hope it will be visible in the viewer then.

dberenbaum: Sure, I kept the old desc here and changed the import desc.

shcheklein (approved):

Issue #51, opened 2021-05-31
Make sed commands for params replacement work across mac/linux


dberenbaum:
`sed -i` has slightly different syntax on mac and linux, causing `generate.sh` to fail on one or the other unless it follows the syntax `sed -i''` suggested in https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux.

PR #53, opened 2021-05-31
Fix #49

### LINKED ISSUE #49: example-get:started: make plot files smaller
> Yeah, technically plotting precision-recall gets a bit weird with fewer points, but as long as we keep a reasonably large number of points, it shouldn’t be a noticeable issue. We can do the same with roc.json

> Agree that we need a good number of data points. It should be enough for Studio to cut the size 2x. And it should not be noticeable for the curve quality (I hope :slightly_smiling_face: ).

> roc.json is very small 56Kb. No need to optimize.

> Oh weird, it looks like sklearn automatically cuts it down to an optimal size for roc but not prc.

> @ivan Let me know if you want me to regenerate it. I don’t know how the additional branch was added, but I can at least take care of master if needed.

cc @dberenbaum @dmpetrov

dberenbaum:
Fix #49 by only using half the points in `prc.json`.

Depends on #51.

### DIFF
--- a/example-get-started/code/src/evaluate.py
+++ b/example-get-started/code/src/evaluate.py
@@ -42,7 +42,7 @@ with open(prc_file, "w") as fd:
         {
             "prc": [
                 {"precision": p, "recall": r, "threshold": t}
-                for p, r, t in zip(precision, recall, prc_thresholds)
+                for p, r, t in list(zip(precision, recall, prc_thresholds))[::2]
             ]
         },
         fd,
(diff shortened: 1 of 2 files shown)


shcheklein:
@dberenbaum merge master? it has some unrelated changes it looks like

example-get-started/code/src/evaluate.py
```diff
@@ -42,7 +42,7 @@
         {
             "prc": [
                 {"precision": p, "recall": r, "threshold": t}
-                for p, r, t in zip(precision, recall, prc_thresholds)
+                for p, r, t in list(zip(precision, recall, prc_thresholds))[::2]
```
shcheklein, line 45: let's put a comment that we are taking half of the points to make it faster for now?

shcheklein (approved):

dberenbaum:
Thanks, please take another look!

PR #73, opened 2021-08-26
Add .gitattributes for syntax-highlighting .dvc and dvc.lock files

### LINKED ISSUE #72: Syntax highlighting for dvc.lock and .dvc files
We should add syntax highlighting for dvc.lock and .dvc files as an yaml file. We can do so by using `.gitattributes` to force Linguist to highlight as such in Github. 

The following text should be kept in `.gitattributes`:
```
*.dvc linguist-language=YAML
dvc.lock linguist-language=YAML
```

See https://github.com/skshetry/example-get-started/blob/e67562ec1983bbcee0e5282227d0d47564c54eb2/dvc.lock for an example.

Until we add the dvc.lock and .dvc to the linguist for syntax highlighting those files automatically.

skshetry:
Closes #72.

### DIFF
--- a/example-dvc-experiments/.gitattributes
+++ b/example-dvc-experiments/.gitattributes
@@ -1,0 +1,2 @@
+*.dvc linguist-language=YAML
+dvc.lock linguist-language=YAML
--- a/example-dvc-experiments/generate.bash
+++ b/example-dvc-experiments/generate.bash
@@ -78,7 +78,6 @@ pip install 'dvc[all]'
 
 git init
 git checkout -b main
-<<<<<<< HEAD:example-dvc-experiments/generate.bash
 cp $HERE/code/README.md "${REPO_PATH}"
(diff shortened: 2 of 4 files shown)


iesahin (approved):
Thank you @skshetry 🙏🏼

example-dvc-experiments/.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.dvc linguist-language=YAML
```
skshetry, line 1: This was renamed to `example-dvc-experiments`, so it's missing the `.gitattributes` copy that I had in https://github.com/iterative/example-repos-dev/commit/a28bbb5bb91694cd9a713038a491f712b3bc4b19.

skshetry: I'll make a separate PR for that.

Issue #34, opened 2021-04-27
get-started-X: Write `generate` scripts for each project


iesahin:
### TODO

- [x] Add an `experiments` branch for various experiments (#38) (Done as a separate repository)
- (CANCEL) Add a `studio` branch to be used as a studio showcase 
- [x] Move current `main` to `pipeline` branch (#39) (Done as a separate repository)
- [x] Add `main` to contain only README (#40) (Cancelled)
- [x] Create Docker containers for each tag
- [x] Update Katacoda GS:Experiments scenario and get rid of `dvc-get-started-mnist`
- [x] Remove model metrics from params
- (MOVE) Keep only the relevant parameters for each repository
- [x] Create two fictional people for each branch, spreading the commits across time
- [x] Add additional plots like confusion matrix
- [ ] Review dvc.org docs examples (besides Get Started) that could use the resulting example repo. Mainly the Experiments docs

### Older Proposal Below. 

We have a new [Get Started project][newgsm] aimed towards experimentation features in DVC 2.0. 

[newgsm]: https://github.com/iterative/dvc-get-started-mnist

A step-by-step approach similar to the current [example-get-started][curgsm] project is necessary for exposition. Creating the project from _development_ sources to provide a clear history, without manual intervention also provides
easier maintenance. 

[curgsm]: https://github.com/iterative/example-get-started

## Tags

- `0-git-init`: Empty Git repository initialized.

- `1-dvc-init`: DVC has been initialized. `.dvc/` with the cache directory created.

- `2-import-data`: Raw data files `data/raw/` downloaded and tracked with DVC using [`dvc import-url`][importurl] from the [dataset-registry][dsr]. First `.dvc` file created.

[importurl]: https://man.dvc.org/import-url
[dsr]: https://github.com/iterative/dataset-registry

- `3-config-remote`: Remote HTTP storage initialized. It's a shared read only storage that contains all data artifacts produced during next steps.

- `4-source-code`: Source code copied from the generation repository. 

- `5-prepare-stage`: Create `dvc.yaml` and the first pipeline stage with [`dvc stage add`][stageadd]. It converts MNIST data from IPX format to NumPy format and stores in `data/prepared/`

[stageadd]: https://man.dvc.org/stage/add

- `6-preprocess-stage`: Create a new stage named `preprocess` to convert the data files produced in the previous stage into a format ready to supply to Keras. 

[...]

PR #75, opened 2021-08-27
add `dvc-checkpoints-mnist`

casperdcl:
- Generate https://github.com/iterative/dvc-checkpoints-mnist


dvc-checkpoints-mnist/branches/full_pipeline/train.py
```diff
@@ -0,0 +1,121 @@
...
+    x = dataset.data.reshape(len(dataset.data), 1, 28, 28) / 255
+    y = dataset.targets
+    return x, y
+
+
+def train(model, x, y, lr, weight_decay):
+    """Train a single epoch."""
+    model.train()
+    criterion = torch.nn.CrossEntropyLoss()
+    optimizer = torch.optim.Adam(model.parameters(), lr=lr, weight_decay=weight_decay)
```
casperdcl, line 46: @dberenbaum isn't this missing `optimizer.load_state_dict()` to ensure the epoch number & weight decay are used correctly?

shcheklein:
Question, @casperdcl @iesahin @dberenbaum - how can we consolidate the repos?

At the moment I'm completely lost in all those repositories. We need to cleanup, come up with the name convention, reuse one or two repos to do experiments, or at least reuse the same code base.

@iesahin what is current status of the experiments-get-started? does it cover checkpoints? what is missing there?

dberenbaum:
My current understanding:

- https://github.com/iterative/dvc-checkpoints-mnist.git is referenced in https://dvc.org/doc/dvclive/user-guide/dvclive-with-dvc and nowhere else. I think it's worth fixing small issues in that repo since it is still being referenced.
- @flippedcoder maintains https://github.com/iterative/checkpoints-tutorial, which is referenced in https://dvc.org/doc/user-guide/experiment-management/checkpoints and nowhere else. I think it's a fork of the first repo, and I'm not sure of the differences. It's also probably worth making updates to this repo and corresponding changes to that tutorial as long as needed.
- @iesahin is working on https://github.com/iterative/get-started-experiments. I hope that this repo or a similar one can eventually be the source for checkpoints docs.

I'm not sure how close we are to switching to the repo of @iesahin. Ideally, I would prefer not to take the time to auto generate the other repos if we think they can be replaced soon.

Sorry @casperdcl if this was unclear. I didn't realize you were planning to try to generate this repo. Maybe we need it if we aren't close with the new repo, but I hope not.

iesahin:
If we have decided on merging `dvc-checkpoints-mnist` and `get-started-checkpoints` and use the latter for all the checkpoints related projects, I can synchronize `get-started-checkpoints` with `get-started-experiments`. `get-started-checkpoints` is already codified in https://github.com/iterative/example-repos-dev/pull/45. It needs some update but I believe it will be quicker than adding a new codification for `dvc-checkpoints-mnist`.

iesahin:
> We need to cleanup, come up with the name convention, reuse one or two repos to do experiments, or at least reuse the same code base.

We visited these topics in the past, e.g. in https://github.com/iterative/example-repos-dev/issues/34. 

- Simple repositories and a single code base seems to contradict at some level. I could use https://github.com/iterative/dvc-get-started for all the experiments / checkpoints / pipelines stuff, but it began to look too complex. I tend to keep per-feature repositories, and they evolve to different code bases. For example, `get-started-pipelines` and `get-started-experiments` started with the same code base but now they are very different. 
- Standardizing on the names may be good. I can update the names [according to the card](https://github.com/orgs/iterative/projects/172#card-67676327) in this week's project: `get-started-experiments` ➡️ `dvc-example-experiments-tensorflow`, `get-started-pipelines` ➡️  `dvc-example-pipelines-tensorflow` and `get-started-checkpoints` ➡️ `dvc-example-checkpoints-tensorflow`. Additionally `dvc-checkpoints-mnist` can be `dvc-example-checkpoints-torch`, `checkpoints-tutorial` can be `dvc-example-checkpoints-torch-tutorial`. `example-get-started` is referenced in too many places, so I believe it's better to keep it intact.

dberenbaum:
@iesahin I'm fine with using `get-started-checkpoints` for future checkpoints-related docs. Do you see anything in `dvc-checkpoints-mnist` that you think needs to be merged or added in `get-started-checkpoints`? I don't think it's necessary to keep anything from `dvc-checkpoints-mnist` unless it's missing from the other repos.

iesahin:
` get-started-experiments`  seems to diverge quite a bit from `get-started-checkpoints`. Instead of modifying the latter, I'm updating `get-started-experiments` to get a new repository `dvc-example-checkpoints-tensorflow`.

iesahin:
Do we need this particular PR to codify `dvc-checkpoints-mnist`? Could we close this? @casperdcl @dberenbaum

dberenbaum:
IMO we don't need to codify `dvc-checkpoints-mnist`.

casperdcl:
I'm not tracking this any more; feel free to take over @iesahin or close it.

Issue #32, opened 2021-04-13
Default branch for example-get-started (and others)


dberenbaum:
My git config sets main as my default branch, which breaks the instructions to publish example-get-started (since I have no master branch to push by default). Ideally, we would update the example-get-started default branch, but we may need to check that we don't break anything in the tutorial. At minimum, we should explicitly checkout a specific branch before publishing.

We should also keep this in mind for other repos in development.

iesahin:
I think I can update the default branch of https://github.com/iterative/example-get-started to `main` in the scripts, but #50 should be done before. (Otherwise custom changes will be overwritten.) In `get-started-experiments`, I used `main` and will continue to do so in others.

shcheklein:
I think this is done.

PR #104, opened 2022-03-30
Use main branch in the example-get-started

shcheklein:
Related #32


iesahin:
I'm building it now. There seems to be other branches and PRs in iterative/example-get-started, they may be affected, are we OK with this? @shcheklein

iesahin (approved):
It's building, although that final message could also be edited.

shcheklein:
> I'm building it now.

Let me rebuild it please, I need to take care of the other branches since it affects Studio. + I want to add the codespaces now that VS Code extension is published.

Thanks, Emre!

Issue #58, opened 2021-06-02
example-get-started: generalize CML workflow


shcheklein:
Right now for some reason (simplicity?) we compare everything with the bigrams experiment.

- we depend on a specific tag name
- when we do a PR to that repo - report name is misleading and not very meaningful

shcheklein:
cc @casperdcl

casperdcl:
> we compare everything with the bigrams experiment

Yes it's for simplicity (made most sense to me at the time - do you have other suggestions?)

> report name is misleading

https://github.com/iterative/example-repos-dev/blob/279b255646d5400cb9ea689302145ec6eb9c5f87/example-get-started/code/.github/workflows/cml.yaml#L15-L25

are you referring to improving L15 above? Or L24?

shcheklein:
> are you referring to improving L15 above? Or L24?

L24. `workspace` term is probably wrong (from the end-user perspective) - we are comparing their commit to some previous attempt. It's not workspace for them.

> do you have other suggestions?

What would be a meaningful CML integration? One example (to my mind). If I do a PR - it should be comparing it with the latest changes in the main branch. When it lands it's merged, or a new commit pushed directly to the main branch - that's a good question. Probably diff doesn't make sense anymore - just show its actual state. Or show it compared to the previous commit.

------

Depending on a hard coded tag name (which is not special and might change) is fragile. And in this case it's not clear why that one specifically.

casperdcl:
I think `dvc metrics diff` uses the term `workspace` though...

daavoo:
Shouldn't the `cml.yaml` workflow also run the train script (i.e. with `pip install -r requirements.txt && dvc repro`) ?? 

I might be missing some context but the repository is used in some DVC Studio demos to showcase how to modify `params` with the `Run new experiment` button. I found it unexpected for the repository to don't actually run the train script with the modified `params` but just to report metrics comparing some hardcoded `refs`.

casperdcl:
Hmm the current code is geared towards you running the experiment locally & using CML for reporting from the CI. The DVC Studio use case would indeed need training as part of the CI too.

PR #105, opened 2022-04-05
get-started: compare with main or previous commit in the CML workflow

shcheklein:
In PRs we now compare metrics and plots with `main`, in the `main` itself we are comparing with `HEAD^1`

Closes https://github.com/iterative/example-repos-dev/issues/58 

Needs some testing, but for simplicity I'm postponing this until I can run everything end to end with all the changes I want to make: add a separate branch for the 4x data, add codespaces, change author names (I believe any human name and face is better than Olivaw)

### DIFF
--- a/example-get-started/code/.github/workflows/cml.yaml
+++ b/example-get-started/code/.github/workflows/cml.yaml
@@ -14,13 +14,14 @@ jobs:
           echo "# CML Report" > report.md
 
           echo "## Plots" >> report.md
-          dvc plots diff bigrams-experiment workspace \
+          dvc plots diff main \
             --show-vega --targets prc.json > vega.json
           vl2svg vega.json prc.svg
           cml publish prc.svg --title "Precision & Recall" --md >> report.md
 
           echo "## Metrics" >> report.md
           echo "### bigrams-experiment → workspace" >> report.md
-          dvc metrics diff bigrams-experiment --show-md >> report.md
+          PREVIOUS_REV=$(if [ ${GITHUB_REF##*/}x = "main"x ]; then echo "HEAD^1"; else echo "main"; fi)
+          dvc metrics diff $PREVIOUS_REV --show-md >> report.md
 
           cml send-comment report.md
--- a/example-get-started/code/.github/workflows/cml.yaml
+++ b/example-get-started/code/.github/workflows/cml.yaml
@@ -21,7 +21,7 @@ jobs:
 
           echo "## Metrics" >> report.md
           if [[ "${GITHUB_REF}" == "refs/heads/main" ]] ; then
-            PREVIOUS_REF='HEAD^1'
+            PREVIOUS_REF='HEAD~1'
           else
             PREVIOUS_REF='main'
           fi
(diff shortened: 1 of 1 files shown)


example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -14,13 +14,14 @@ jobs:
...
-          dvc plots diff bigrams-experiment workspace \
+          dvc plots diff main \
             --show-vega --targets prc.json > vega.json
           vl2svg vega.json prc.svg
           cml publish prc.svg --title "Precision & Recall" --md >> report.md
 
           echo "## Metrics" >> report.md
           echo "### bigrams-experiment → workspace" >> report.md
-          dvc metrics diff bigrams-experiment --show-md >> report.md
+          PREVIOUS_REV=$(if [ ${GITHUB_REF##*/}x = "main"x ]; then echo "HEAD^1"; else echo "main"; fi)
```
iesahin, line 24: How about:
```
if [[ "${GITHUB_REF}" == "refs/heads/main" ]] ; then
   PREVIOUS_REF='HEAD^1'
else
   PREVIOUS_REF='main'
fi
```

shcheklein: done, thanks, Emre

example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -14,13 +14,18 @@ jobs:
...
+          dvc plots diff main \
             --show-vega --targets prc.json > vega.json
           vl2svg vega.json prc.svg
           cml publish prc.svg --title "Precision & Recall" --md >> report.md
 
           echo "## Metrics" >> report.md
-          echo "### bigrams-experiment → workspace" >> report.md
-          dvc metrics diff bigrams-experiment --show-md >> report.md
+          if [[ "${GITHUB_REF}" == "refs/heads/main" ]] ; then
+            PREVIOUS_REF='HEAD^1'
```
casperdcl, line 24: I tend to prefer `~` because it work on Windows & Linux (`^` is an escape char on Win)

```suggestion
            PREVIOUS_REF='HEAD~1'
```

casperdcl (commented):
> any human name and face is better than Olivaw

I guess this means borrowing from the previous commit? https://github.com/iterative/magnetic-tiles-defect/blob/c449439a9d3b04d14a95bbc39b567c6d4d7ae5cd/.github/workflows/1-experiment.yaml#L45-L49

casperdcl (approved):

example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -14,13 +14,18 @@ jobs:
           echo "# CML Report" > report.md
 
           echo "## Plots" >> report.md
-          dvc plots diff bigrams-experiment workspace \
+          dvc plots diff main \
```
pmrowla, line 17: Why is the `dvc plots diff` target different than `dvc metrics diff`? It seems like they should either both be comparing against `main` or both be comparing against `HEAD~1`.

(from https://github.com/iterative/dvc/issues/7578#issuecomment-1102545000)

PR #106, opened 2022-04-07
get-started: codify 4x data branch

### LINKED ISSUE #50: example-get-started: codify the branch with CML report and larger dataset
As we use now project as a demo for studio, we added the new branch and PR:

https://github.com/iterative/example-get-started/pull/8

- we need to add commands to the script to generate this branch
- we need to add commands to generate this PR
- make sure that we drop the previous one
- make sure that CML reports are being generated

shcheklein:
Fixes #50 

+ Use real author names, it looks better in Studio

Still need to generate the PR + launch action on the last commit. `gh` app or something like that is needed.

### DIFF
--- a/example-get-started/generate.sh
+++ b/example-get-started/generate.sh
@@ -28,8 +28,8 @@ tick(){
   export GIT_COMMITTER_DATE=${TAG_TIME}
 }
 
-export GIT_AUTHOR_NAME="Olivaw Owlet"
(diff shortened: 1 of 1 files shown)


iesahin (approved):
👍🏼

iesahin:
Should I build and push this version?

shcheklein:
@iesahin thanks. no worries, a bit too early for that ... let me do some actual testing ... I'll publish it when time comes.

PR #107, opened 2022-04-09
Get-started: fix ML workflow

shcheklein:
* A few fixes are needed to run the new CML workflow
* Add a few commands on how to generate PR for the try large dataset branch


jorgeorpinel:
@casperdcl @0x2b3bfa0 could you PTAL?

shcheklein:
@casperdcl - this is a quick fix, please take a look when you have time :)

example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -11,21 +11,21 @@ jobs:
           REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           cml ci
-          echo "# CML Report" > report.md
+          if [ "$GITHUB_REF" = "refs/heads/main" ] ; then
+            PREVIOUS_REF='HEAD~1'
+          else
+            PREVIOUS_REF='main'
+          fi
```
0x2b3bfa0, line 18:
```suggestion
          if [ "$GITHUB_REF" = refs/heads/main ]; then
            PREVIOUS_REF=HEAD~1
          else
            PREVIOUS_REF=main
          fi
```

Optional nitpick: remove unnecessary quotes and space before semicolon.

casperdcl (approved):
lgtm - same fix as earlier but moved higher up.

Wonder if this can be avoided by changing the default behaviour of `dvc metrics diff` to automatically compare with `(REF_unspecified_on_CLI && $GIT_REF == HEAD) ? HEAD~1 : $MAIN_BRANCH`

shcheklein:
I don't like the fix (I prefer to quote string, literals - in more or less complicated script it's easier to read and more consistent), but let's move forward, not the biggest problem ... fingers crossed that it's not breaking it (since we are not testing this version)

PR #108, opened 2022-04-10
get-started: add codespaces support with VS Code extension

shcheklein:


example-get-started/generate.sh
```diff
@@ -219,7 +218,7 @@ git push --force origin --tags
 
 Run these to drop and then rewrite the experiment references on the repo:
 
-git ls-remote origin "refs/exps/*" | awk '{print \$2}' | xargs -n 1 git push -d origin
+git ls-remote origin "refs/exps/*" | awk '{print \\$2}' | xargs -n 1 git push -d origin
```
jorgeorpinel, line 221: Is this right? It's different than https://github.com/iterative/example-repos-dev/pull/106/files#diff-f72bf81b747918d477ba9f4d2d4251111fbf04faf815d95716c13a093314a372L204-R222 . Maybe merge `main` to this branch instead.

jorgeorpinel (commented):
Looks good but we'd need to generate this and upload to Github to make sure... ⌛

shcheklein:
@jorgeorpinel yep, it's done here https://github.com/iterative/example-get-started/pull/18

jorgeorpinel (approved):
Ah OK (I almost made my own ha). Yup, it works!

image


Should we update the RADME to explain how to use it too? (Plus it needs a few other small updates probably).

jorgeorpinel:
One thing I'm not sure about is the `sshd` process that's automatically launched though. I don't think that's normal but can't tell where it comes from

image

jorgeorpinel:
> it seems to work!

Well, one more thing I don't see in based on https://github.com/iterative/example-get-started/pull/18 is all the [experiments that are run](https://github.com/iterative/example-repos-dev/blob/master/example-get-started/generate.sh#L163-L169) in this generation script though. Shouldn't those be visible by default?

jorgeorpinel:
> One thing I'm not sure about is the sshd process that's automatically launched

p.s. this only happened when I created the Codespace. No longer happening after stopping and restarting 🤷🏼

jorgeorpinel:
> one more thing I don't see ... is all the [experiments](https://github.com/iterative/example-repos-dev/blob/master/example-get-started/generate.sh#L163-L169) ...

Ah never mind, by default it only shows experiments from that text branch (`codespaces`) and there's none. `dvc exp show --rev master` shows the one experiment that exists in that repo's master branch (I assume it was generated with an old script).

casperdcl (commented):
qq - should DVCLive & CML be added to the container?

jorgeorpinel:
> qq - should DVCLive & CML be added to the container?

They're not used in the get-started example repo are they?

jorgeorpinel:
@shcheklein @iesahin should we generate and deploy this to https://github.com/iterative/example-get-started ? For a final check -- and probably update the README with basic instructions and links to Codespaces docs.

shcheklein:
@jorgeorpinel I'll handle this, guys. CML workflow is not working atm (due to some recent changes that GH made) - we need to wait for a bit. Also, this project is used in Studio - we should be careful updating it.

PR #109, opened 2022-04-19
Add mlem get started

mike0sv:


mike0sv:
Converting this to draft since actual docs are not live on mlem.ai, but any feedback is still welcomed

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+
+cp $HERE/code/src/evaluate.py .
+python evaluate.py
+git add metrics.json evaluate.py
+tick
+git commit -m "Evaluate model"
+git tag -a "4-eval-stage" -m "Metrics calculated"
+
+git rm -r --cached .mlem/model/rf .mlem/dataset/train.csv .mlem/dataset/test_x.csv .mlem/dataset/test_y.csv
+git commit -m "stop tracking data"
```
shcheklein, line 100: why is it called "stop tracking data"?

shcheklein: do we need this extra commit?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+
+dvc init
+dvc remote add myremote azure://example-mlem --default
+mlem config set default_storage.type dvc
+python train.py
+dvc add .mlem/model/rf .mlem/dataset/train.csv .mlem/dataset/test_x.csv .mlem/dataset/test_y.csv
```
shcheklein, line 106: that's not very common for DVC to track each independently (esp dataset) - is it intended?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+
+Run these commands to force push it:
+
+cd build/example-mlem-get-started
+git remote add origin  https://github.com/iterative/example-mlem-get-started
+git push --force origin main
+git push --force origin --tags
+cd ../../
+
+Run these to drop and then rewrite the experiment references on the repo:
```
shcheklein, line 142: you don't need this anymore I believe

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+unset GIT_COMMITTER_DATE
+unset GIT_AUTHOR_NAME
+unset GIT_AUTHOR_EMAIL
+unset GIT_COMMITTER_NAME
+unset GIT_COMMITTER_EMAIL
+
+echo "`cat <<EOF-
+
+The Git repo generated by this script is intended to be published on
+https://github.com/iterative/example-mlem-get-started. Make sure the Github repo
```
shcheklein, line 126: please, let's try to keep it formatted <= 80

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+git add metrics.json evaluate.py
+tick
+git commit -m "Evaluate model"
+git tag -a "4-eval-stage" -m "Metrics calculated"
+
+git rm -r --cached .mlem/model/rf .mlem/dataset/train.csv .mlem/dataset/test_x.csv .mlem/dataset/test_y.csv
+git commit -m "stop tracking data"
+
+dvc init
+dvc remote add myremote azure://example-mlem --default
```
shcheklein, line 103: do we need to use `--local` here and add `-d` http storage on top so that model is accesible?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+source $BUILD_PATH/.venv/bin/activate
+
+REPO_PATH="$HERE/build/$REPO_NAME"
+
+if [ -d "$REPO_PATH" ]; then
+  echo "Repo $REPO_PATH already exists, please remove it first."
+  exit 1
+fi
+
+TOTAL_TAGS=15
```
shcheklein, line 37: change this?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+
+pushd $BUILD_PATH
+if [ ! -d "$BUILD_PATH/.venv" ]; then
+  exit 1
+  virtualenv -p python3 .venv
+  export VIRTUAL_ENV_DISABLE_PROMPT=true
+  source .venv/bin/activate
+  echo '.venv/' > .gitignore
+  pip install gitpython
+  pip install "git+https://github.com/iterative/mlem#egg=mlem[all]" --use-deprecated=legacy-resolver
```
shcheklein, line 23: `--use-deprecated=legacy-resolver` why is it needed - out of curiosity? is it some known bug /problem?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+git tag -a "1-mlem-init" -m "MLEM initialized."
+
+
+cp $HERE/code/src/requirements.txt .
+cp $HERE/code/src/prepare.py .
+git add .
+python prepare.py
+git add .mlem/dataset
+tick
+git commit -m "Create data stage"
```
shcheklein, line 80: in this case it can be as simple as "Add data" ? (no need to use "stage")

shcheklein: ah, I see, it actually prepares data ... seems a bit too complicated ... should we just pre-bundle those files with code?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,156 @@
...
+
+mlem init
+tick
+git add .mlem/config.yaml
+git commit -m "Initialize MLEM project"
+git tag -a "1-mlem-init" -m "MLEM initialized."
+
+
+cp $HERE/code/src/requirements.txt .
+cp $HERE/code/src/prepare.py .
```
shcheklein, line 75: If you don it similar to DVC get-started you need to bundle this code and share from S3 so that people could curl or wget it

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+## Existing stages
+
+This project with the help of the Git tags reflects the sequence of actions that
+are run in the MLEM [get started](https://mlem.ai/doc/get-started) guide. Feel
+free to checkout one of them and play with the MLEM commands having the
+playground ready.
+
+[comment]: <> (TODO)
```
shcheklein, line 36: resolve this, not sure this is needed at all in this case - it's simpler than the DVC's project

shcheklein (changes requested):
Thanks a lot for codifying this 🙏 

I put some comments here and there to address, but also:

- should we create a branch in the regular example-get-started that converts models into mlem and uses gto + deploys it. That would be cool.
- in this example - should we have something about client, deployment, gto? Just saving model might be not very interesting (not clear why would I want this at all)

example-mlem-get-started/generate.sh
aguschin: Yes, it's a known issue: https://github.com/iterative/mlem/issues/140

example-mlem-get-started/generate.sh
mike0sv: there are .mlem files in this dir. Can we do something like `echo *.mlem > .dvcignore && dvc add data`?

mike0sv: the correct pattern was found and it is `/**/?*.mlem`. Btw, does gitignore support ! negation?

example-mlem-get-started/generate.sh
mike0sv: we want to show how to save data also. So reading a file, and then re-saving it to mlem is even more compicated

mike0sv:
1. Yes, but this could wait until MLEM release I guess
2. I am planning to add something like that later, this was for first two sections of get-started https://github.com/iterative/mlem.ai/pull/43/files#diff-1cd29580767fad4637f461df8460373ba6ec523dc598d7d3550657ed56cab101

example-mlem-get-started/generate.sh
mike0sv: It's faster and more importantly works (new resolver breaks until we resolve https://github.com/iterative/mlem/issues/140)

example-mlem-get-started/generate.sh
mike0sv: this does not seem to work, since we cant `dvc add .mlem/dataset/` since we lose `.mlem` files, even though they ignored. So going with `dvc add .mlem/dataset/*.csv`. Would be great to have some way to do "add everything except dvc-ignored"

example-mlem-get-started/generate.sh
shcheklein: but I don't think we need an extra commit ... could you clarify `this one is suggested by DVC itself`?

example-mlem-get-started/generate.sh
shcheklein: > Btw, does dvcignore support ! negation?

yes, it does support it

it should be working, otherwise it's a bug to my mind

example-mlem-get-started/generate.sh
shcheklein: yep, but does it mean that we will recommend the same in the Install guide and README? looks a bit weird ... (not critical, but I would try to address this before the release if possible)

example-mlem-get-started/generate.sh
shcheklein: raw.github.com works as well, I don't see any particular problems with that .. I would bundle though everything as a single tar (or may github has a way to download revisions as a tar with curl, I don't know)

shcheklein:
Thanks, @mike0sv for addressing some things. I left a bit more comments, but nothing major.

example-mlem-get-started/generate.sh
mike0sv: image
without those two lines. But it works without commit, I checked. Just thought commit is necessary since DVC suggested it

example-mlem-get-started/generate.sh
mike0sv:
```
$ cat .dvcignore 
/**/*.mlem
!/.mlem/
(.venv) (mlem) MacBook-Pro-5:example-mlem-get-started mike0sv$ dvc check-ignore .mlem -d
.dvcignore:2:!/.mlem/   .mlem
$ cat .dvcignore 
/**/*.mlem
!/.mlem/
$ dvc check-ignore .mlem -d
.dvcignore:2:!/.mlem/   .mlem
```
I was trying to make an exception for `.mlem` dir in a root, but this is what I've got

example-mlem-get-started/generate.sh
mike0sv: we'll deal with it before release

example-mlem-get-started/generate.sh
mike0sv: https://github.com/iterative/example-mlem-get-started/archive/refs/tags/5-switch-to-dvc.tar.gz for example

example-mlem-get-started/generate.sh
shcheklein: q - why do you need an exception like this? (curious)

does the same combination work for gitignore?

example-mlem-get-started/generate.sh
shcheklein: it contains some extra files that are not needed initially though?

example-mlem-get-started/generate.sh
mike0sv: My goal is to exclude all `*.mlem` files, but not `.mlem` dir in repo root. But simple `/**/*.mlem` adds `.mlem` dir to ignore list. With git it works

example-mlem-get-started/generate.sh
mike0sv: it contains whole repo, yes. But I don't see any point of maintaining them separately, since they are very small and easy and also will be available in the docs themselves

shcheklein: it's not about the size, it's about the workflow. In the dvc.org get started you would have a command:

```
wget code.zip
unzip code.zip
git add -a -m "add code"
```

https://dvc.org/doc/start/data-pipelines#expand-to-download-example-code

including any extra files there would confuse and ruin the flow of the document. 

so, your call here - if you want something similar to DVC then consider creating a clean code package.

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,149 @@
+#!/bin/bash
+# See https://dvc.org/doc/tutorials/get-started
```
aguschin, line 2:
```suggestion
# See https://mlem.ai/doc/get-started
```

example-mlem-get-started/generate.sh
aguschin: also, should MLEM be installed from git here? Not from PyPi?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,149 @@
...
+BUILD_PATH="$HERE/build"
+
+pushd $BUILD_PATH
+if [ ! -d "$BUILD_PATH/.venv" ]; then
+  exit 1
+  virtualenv -p python3 .venv
+  export VIRTUAL_ENV_DISABLE_PROMPT=true
+  source .venv/bin/activate
+  echo '.venv/' > .gitignore
+  pip install gitpython
```
aguschin, line 22: Why you need to install gitpython?

example-mlem-get-started/generate.sh
mike0sv: we want git since we may want to prepare example repos before new release

example-mlem-get-started/generate.sh
mike0sv: not sure)))))

shcheklein: this is related to the way DVC calculates it's version in the dev env - https://dvc.org/doc/install/pre-release (add hash to the end of the version). If MLEM does the same (and it should be) then you might also need this. I would check if DVC has changed this mechanism though.

shcheklein: thanks, created https://github.com/iterative/example-repos-dev/pull/111 and removed it from the docs (pending redeploy)

example-mlem-get-started/generate.sh
shcheklein: thinking second time - I remember now why I didn't use GH (most likely) - I wanted to keep the command as short and nice as possible `wget https://code.dvc.org/get-started/code.zip` and zip instead of tar

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,39 @@
...
+$ gh repo fork https://github.com/iterative/example-mlem-get-started
+$ cd example-mlem-get-started
+```
+
+Now let's install the requirements. But before we do that, we **strongly**
+recommend creating a virtual environment with a tool such as
+[virtualenv](https://virtualenv.pypa.io/en/stable/):
+
+```console
+$ python3 -m venv .env
```
shcheklein, line 29: as far as I know it doesn't install wheels (slow downs and / or makes installing the requirements process slow / broken)

example-mlem-get-started/generate.sh
shcheklein: is it fixed?

example-mlem-get-started/code/src/train.py
```diff
@@ -0,0 +1,23 @@
...
+        n_jobs=2,
+        random_state=42,
+    )
+    rf.fit(data, y)
+
+    save(
+        rf,
+        "rf",
+        sample_data=data,
+        description="Random Forest Classifier",
```
shcheklein, line 18: make more meaningful description

example-mlem-get-started/code/src/train.py
```diff
@@ -0,0 +1,23 @@
...
+def main():
+    data, y = load_iris(return_X_y=True, as_frame=True)
+    rf = RandomForestClassifier(
+        n_jobs=2,
+        random_state=42,
+    )
+    rf.fit(data, y)
+
+    save(
+        rf,
```
shcheklein, line 15: should we call it `model` for simplicity?

shcheklein:
@iterative/mlem hey folks, any progress on this?

mike0sv:
Since it's already live I think we should go ahead and merge this

shcheklein:
@mike0sv @aguschin let's cleanup it first please, it has `fixture-model-registry` for example now besides the `get-started` project.

example-mlem-get-started/code/src/train.py
aguschin: Don't think it matter that much. And we have docs in place, that rely on this repo, so let's not change this.

example-mlem-get-started/code/src/train.py
aguschin: Let's keep this. I don't think it's a problem and we use this in docs already.

aguschin (approved):
I think we can merge this now. @shcheklein?

example-mlem-get-started/code/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "experiments-get-started",
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
+  "postCreateCommand": "pip3 install --user mlem"
```
shcheklein, line 6: should we also install the requirements?

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,39 @@
+# MLEM Get Started
+
+This is an auto-generated repository for use in MLEM
+[Get Started](https://mlem.ai/doc/get-started). It is a step-by-step quick
+introduction into basic MLEM concepts.
+
+The main branch contains only the code needed for the tutorial. Results of each step can be found in tags, and resulting state in correctponding branches.
```
shcheklein, line 7: please apply the standard formatting (80 symbols, etc) ... I need to check if pre-commit is setup for this repo

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+```console
+$ python3 -m venv .env
+$ source .env/bin/activate
+$ pip install -r requirements.txt
+```
+
+## Existing stages
+
+This project with the help of the Git tags reflects the sequence of actions that
```
shcheklein, line 36: Git tags - let's make it a link to https://github.com/iterative/example-mlem-get-started/tags 

Also, it looks it's a mess now - I see a few tags that start with the same number

example-mlem-get-started/code/src/requirements.txt
```diff
@@ -0,0 +1,8 @@
+pandas
+scikit-learn
+scipy
+dvc[azure,s3]
```
shcheklein, line 4: installing DVC like this will contradict the devcontainer that installs it globally - btw, why do we need it?

example-mlem-get-started/code/src/requirements.txt
```diff
@@ -0,0 +1,8 @@
...
+fastapi
+uvicorn
+docker
+mlem
```
shcheklein, line 8: in the devcontainer we also install it, should we just install the requirements there?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,210 @@
...
+  git clone https://github.com/iterative/mlem.git
+  pip install -e ./mlem
+fi
+popd
+
+source $BUILD_PATH/.venv/bin/activate
+
+REPO_PATH="$HERE/build/$REPO_NAME"
+
+#if [ -d "$REPO_PATH" ]; then
```
shcheklein, line 41: dead code

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,210 @@
...
+echo ".venv" > .gitignore
+git add .
+tick
+git commit -m "Initialize Git repository"
+git tag -a "0-git-init" -m "Git initialized."
+
+
+################## MLEM
+
+git branch simple
```
shcheklein, line 83: Q: so, do we use main or simple? do we use main at all?

shcheklein:
@mike0sv @aguschin did the final review, put some minor questions / comments / bugs - PTAL

example-mlem-get-started/code/README.md
aguschin: It works currently. Do we need to investigate something?

example-mlem-get-started/code/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "experiments-get-started",
```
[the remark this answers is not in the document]
aguschin: Done!

example-mlem-get-started/code/README.md
aguschin: Done.

example-mlem-get-started/code/README.md
aguschin: Sorry, which ones do start with the same number?
image

example-mlem-get-started/code/src/requirements.txt
aguschin: We need DVC here because in one branch (`dvc`) we storing models with DVC. We don't cover this in GS, but we use this in UC: https://mlem.ai/doc/use-cases/dvc

aguschin: I need to check whether this actually breaks something or if it's ok to have.

example-mlem-get-started/code/src/requirements.txt
aguschin: Removed the one from devcontainer.

example-mlem-get-started/generate.sh
aguschin: We use `main` later on L140 as a base for `dvc` branch. So `main` is the common beginning for both non-DVC and DVC cases.

example-mlem-get-started/code/src/requirements.txt
aguschin: Ok, I checked and it doesn't seem to cause any problems.

example-mlem-get-started/code/README.md
shcheklein: no, it's fine (I would not recommend using it, and we don't, but it's simpler and it's not critical). The only thing is - please fix the link that leads to the virtualenv above

example-mlem-get-started/code/README.md
shcheklein: yes, please disregard that comment about the number

example-mlem-get-started/code/src/requirements.txt
shcheklein: Yep, I still think it's better in the use case (which is not actually even a use case, but a tutorial if we apply the DVC.org terminology) to mention the link to the DVC install page. People can decide how to install it. Unless mlem depends on it (but them there should be something like mlem[dvc] or something)

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,209 @@
...
+git tag -a "4-pack" -m "Pip package config added"
+
+mlem declare env heroku staging
+mlem declare deployment heroku myservice -c app_name=example-mlem-get-started-app -c model=rf -c env=staging
+git add .mlem
+tick
+git commit -m "Add env and deploy meta"
+git tag -a "5-deploy-meta" -m "Target env and deploy meta added"
+
+if heroku apps:info example-mlem-get-started-app; then
```
shcheklein, line 127: now, we need to change it?

aguschin: Yes, doing that in https://github.com/iterative/mlem.ai/pull/131

example-mlem-get-started/code/src/requirements.txt
aguschin: Ok, it's here https://github.com/iterative/mlem.ai/pull/132
Could you please review?

example-mlem-get-started/code/README.md
aguschin: > would be nice to explain that there are two branches I think

Done :)

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,44 @@
...
+To be able to commit and push your results please fork and then clone this repo
+
+```console
+$ gh repo fork https://github.com/iterative/example-mlem-get-started
+$ cd example-mlem-get-started
+```
+
+Now let's install the requirements. But before we do that, we **strongly**
+recommend creating a virtual environment with a tool such as
+[virtualenv](https://virtualenv.pypa.io/en/stable/):
```
shcheklein, line 31: we mention it, but we don't use (instead we use the Python's built-in venv module). As we agreed it's fine, but we should remove this link then

example-mlem-get-started/code/src/requirements.txt
shcheklein: done, I meant though that I would not include DVC as a dependency ... I would mention in the tutorial something like: `We assume that DVC is installed ... use this page to install it ...`

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,44 @@
+# MLEM Get Started
+
+This is an auto-generated repository for use in MLEM
+[Get Started](https://mlem.ai/doc/get-started). It is a step-by-step quick
+introduction into basic MLEM concepts.
+
+- The `main` branch contains only the code needed to start the Get Started. This
```
shcheklein, line 7: That's quite unusual - downside is that people who come to the repo won't see anything in it. This is not good to my mind. What is the purpose of keeping the main like this at all if we have the corresponding tags? Can it be done this way:

- `simple` becomes `main`
- `dvc` - can it be done in way that builds on top of simple - we can make a PR from `dvc` to `main` (ex-simple) to DVC-ify the project?

example-mlem-get-started/generate.sh
```diff
@@ -0,0 +1,209 @@
...
+
+To create it with https://cli.github.com/, run:
+
+gh repo create iterative/example-mlem-get-started --public \
+     -d "Get Started MLEM project" -h "https://mlem.ai/doc/get-started"
+
+Run these commands to force push it:
+
+cd build/example-mlem-get-started
+git remote add origin  https://github.com/iterative/example-mlem-get-started
```
shcheklein, line 200: extra space

+ let's rename `iterative` into `shcheklein` (or any other name) for example ... I found that people copy-paste and push this sometimes by mistake / too early

shcheklein:
@aguschin thanks! we are getting there :) I still have some questions though. Please see some unresolved comments.

example-mlem-get-started/code/src/requirements.txt
aguschin: Then UC won't work for those who don't have DVC installed. Alternatively, we could ask for DVC to be installed globally and remove it from `requirements.txt`. Although I'm not too fond of this - I would try to keep all installables inside the venv we create, so the user could remove it later. Maybe DVC case is an exception here, don't know.

example-mlem-get-started/code/README.md
aguschin: https://github.com/iterative/mlem.ai/pull/132#issuecomment-1185452401

example-mlem-get-started/code/README.md
aguschin: Did that and committed changes. Please see https://github.com/iterative/mlem.ai/pull/132 for PR updating documentation.

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,40 @@
+# MLEM Get Started
+
+This is an auto-generated repository for use in MLEM
+[Get Started](https://mlem.ai/doc/get-started). It is a step-by-step quick
+introduction into basic MLEM concepts.
+
+- The `main` branch contains the results of following the Get Started. Results
+  of each step can be found in git tags.
```
shcheklein, line 8: minor: git -> Git

example-mlem-get-started/code/README.md
```diff
@@ -0,0 +1,40 @@
...
+[example-repos-dev](https://github.com/iterative/example-repos-dev).
+
+## Installation
+
+Python 3.7+ is required to run code from this repo.
+
+To be able to commit and push your results please fork and then clone this repo
+
+```console
+$ gh repo fork https://github.com/iterative/example-mlem-get-started
```
shcheklein, line 22: It's really bad practice to depend on `gh`, we should prefer using vanilla `git` commands as were discussing before.

example-mlem-get-started/code/src/requirements.txt
shcheklein: do we need to update mlem -> mlem[dvc-s3, dvc-azure]?

shcheklein: and remove `dvc` itself?

and even in this case not a big fan of this tbh, we should have a way to install `dvc-core-s3` or something, so that CLI is not installed

not a good practice to have CLI tools as dependencies, it smells

shcheklein (dismissed):
Some minor comments left and I didn't try to run it.

Good progress, we are almost there.

Overall, still look too complicated for the MLEM get started. I think as we come to the updates on docs we can optimize this further.

aguschin:
Thanks for the feedback. Fixed those @shcheklein.

shcheklein:
@aguschin one unresolved item left - https://github.com/iterative/example-repos-dev/pull/109/files#r919031250 - up to you to address or merge! Good stuff.

example-mlem-get-started/code/src/requirements.txt
aguschin: > My point is that it's rare you would include dvc as a requirement unless you use it as a library (Python API).

Ok, IIUC, we could keep this as is, cause MLEM calls DVC API and DVC should be installed with `pip`. We already state that DVC is installed from pip. As we install all other packages without `--user`, I think we should install DVC like this as well. Do I misunderstand something?

aguschin: Btw, from the other hand we have `fastapi` and `docker` installed with pip, although they may be installed as MLEM dependencies 🤔 Not sure what's the right approach here, but I think we should be consistent?

aguschin: @shcheklein, maybe I misunderstood you here? Do you maybe suggest to `pip install mlem dvc[azure,s3]` first and then `pip install -r requirements.txt` that have all other dependencies in it?

shcheklein: > , I guess your last comment was about replacing dvc[azure,s3] with mlem[dvc,dvc-azure,dvc-s3], not about this discussion itself. Forgot to do that 😅

yes! sorry about the confusion

if fastapi / docker are used only by MLEM I would definitely install them via MLEM also!

aguschin:
All changes addressed. I've just updated https://github.com/iterative/example-mlem-get-started

Now merging https://github.com/iterative/mlem.ai/pull/132 and https://github.com/iterative/example-repos-dev/pull/109

Thanks for your help and feedback @jorgeorpinel and @shcheklein !

PR #110, opened 2022-04-21
Add GTO get started

aguschin:
Ok, I created PR to `example-repos-dev`, but in the process I think we losed the great way for users to regenerate the repo themselves (this is mode important compared to DVC or MLEM). Earlier it was like this: fork, clone, checkout to dev, run bash script. Now you can also do that, but you need to clone `example-repos-dev`, which is another repo, and then monkeypatch `generate.sh`, then deal with `gh` and etc. It's acceptable for us, but not so easy for users. I will consider creating self-contained example repo which could be used in old scenarios.


shcheklein:
> but you need to clone example-repos-dev, which is another repo

this repo is not meant to be used by users / readers. This repo is done for us to help us maintain tutorials / example repos. E.g. if I want to change something in the middle of the example repo history, update tools, updates all .dvc files in the history to the newer version, etc - this repo help a lot with this.

In case of GTO this repo could be used to generate the `self-contained example repo` and then text tutorial can show users what do with it. Essentially this repo is for _us_ to help us maintain and keep up to date the _self-contained example repo_ in this case.

Or it can be that you decide to all the steps for users and make it the final result - that's also fine, our docs are written in way that users can take the final repo result if they want, but they could also go thought the process from the very beginning (git init, dvc init, add code, etc) or from a tag in the middle (and we provide the list of tags).

example-gto/code/README.md
```diff
@@ -0,0 +1,30 @@
+# GTO Get Started
+
+This is an auto-generated repository for use in [GTO Get Started](https://github.com/iterative/gto/blob/main/README.md). It is a step-by-step quick
```
shcheklein, line 3: keep it formatted (we should enable prettier in this repo?)

example-gto/code/.github/workflows/gto-act-on-tags.yml
```diff
@@ -0,0 +1,46 @@
+name: Act on git tags produced by GTO
```
shcheklein, line 1: what is the purpose of this action? let's say we would be actually using GTO, would we need all the things that this action does? 

can we skip full unshallow clone (it's not good if ppl will start copy-pasting this w/o need), running gto show, etc -- and get to `gto check-ref $GITHUB_REF` part that would do something with that REF? e.g runs a model or something?

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+BUILD_PATH="$HERE/build"
+
+mkdir $BUILD_PATH
+pushd $BUILD_PATH
+if [ ! -d "$BUILD_PATH/.venv" ]; then
+  virtualenv -p python3 .venv
+  export VIRTUAL_ENV_DISABLE_PROMPT=true
+  source .venv/bin/activate
+  echo '.venv/' > .gitignore
+  pip install gitpython
```
shcheklein, line 29: is it needed here? :)

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+
+HERE="$( cd "$(dirname "$0")" ; pwd -P )"
+USER_NAME="iterative"
+REPO_NAME="example-gto"
+
+BUILD_PATH="$HERE/build"
+
+mkdir $BUILD_PATH
+pushd $BUILD_PATH
+if [ ! -d "$BUILD_PATH/.venv" ]; then
```
shcheklein, line 24: it's better to fail the whole script if BUILD PATH exists

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+  virtualenv -p python3 .venv
+  export VIRTUAL_ENV_DISABLE_PROMPT=true
```
shcheklein, line 26: what does it do?

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+  pip install gitpython
+  pip install gto
+fi
+popd
+
+source $BUILD_PATH/.venv/bin/activate
+
+REPO_PATH="$HERE/build/$REPO_NAME"
+
+if [ -d "$REPO_PATH" ]; then
```
shcheklein, line 38: this should probably go before ... I would keep e2e reproducible, so that builds do not affect each other in any way

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+export GIT_AUTHOR_EMAIL="1aguschin@gmail.com"
+export GIT_COMMITTER_NAME="$GIT_AUTHOR_NAME"
+export GIT_COMMITTER_EMAIL="$GIT_AUTHOR_EMAIL"
+
+
+
+mkdir -p $REPO_PATH
+pushd $REPO_PATH
+
+git init -b main
```
shcheklein, line 65: hmm ... is `-b main` needed? (we are not using it in other places?)

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+cp -r $HERE/code/ .
+git add .
+tick
+git commit -m "Initialize Git repository with CI workflow"
+
+if $PUSH; then
+  # remove GH Actions workflows
+  gh api repos/$USER_NAME/$REPO_NAME/actions/runs \
```
shcheklein, line 73: this should go at the end of the script ideally, we should not be changing the remote repo in the middle. Can we do it at the end?

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+    --paginate -q '.workflow_runs[] | "\(.id)"' | \
+    xargs -n1 -I % gh api --silent repos/$USER_NAME/$REPO_NAME/actions/runs/% -X DELETE
+  # add remote
+  git remote add origin https://github.com/$USER_NAME/$REPO_NAME
+  # remove all tags from remote
+  git ls-remote --tags origin | awk '/^(.*)(\s+)(.*[a-zA-Z0-9])$/ {print ":" $2}' | xargs git push origin
+fi
+
+echo "Fix env"
+pip freeze > requirements.txt
```
shcheklein, line 83: requirements file should be part of the code folder, `pip freeze` will produce a lot of things (probably all gto dependencies)

example-gto/generate.sh
```diff
@@ -0,0 +1,186 @@
...
+echo "Create new models"
+mkdir models
+echo "1st version" > models/churn.pkl
+git add models requirements.txt
+tick
+git commit -am "Create models"
+
+gto annotate churn --type model --path models/churn.pkl --must-exist
+gto annotate segment --type model --path s3://mycorp/proj-ml/segm-model-2022-04-15.pt
```
shcheklein, line 92: can we make it realistic - actual existing path with some actual mode?

shcheklein:
@aguschin is it still relevant?

aguschin:
@shcheklein, yes. I'll address this after MLEM release.

shcheklein:
@aguschin closing this? is it still relevant?

aguschin:
@shcheklein, yes, it is. This generates https://github.com/iterative/example-gto

shcheklein:
@aguschin let's finalize and merge it properly then please?

example-gto/generate.sh
aguschin: https://virtualenv.pypa.io/en/16.7.9/reference.html#envvar-VIRTUAL_ENV_DISABLE_PROMPT

example-gto/generate.sh
aguschin: Thanks for the comment, made changes.

example-gto/generate.sh
aguschin: Yes, otherwise it's generated with `master` branch as default one. Probably it's fixed some other way in DVC get started.

example-gto/generate.sh
aguschin: If you run the script with --push, it pushes everything commit by commit. It's important since the `gto show` that runs in CI will give you different results if the repo was pushed already, compared to pushing it "step by step".

aguschin: This specific line removes old CI runs from the example repo so you won't see them after you force push new generated repo.

example-gto/generate.sh
aguschin: Yes, but that I would skip this for now since you cannot do anything with that model anyway, at least within GTO. For example purposes it's fine. Do you disagree?

example-gto/code/.github/workflows/gto-act-on-tags.yml
aguschin: > Yes, I think we can add a usage example here

Checked, currently all models are just some non-existing references. To actually add a usage example, you would need to (1) create real models in buckets, (2) based on check-ref results actually apply those models.

I agree that this is important both for demo and for understanding whether we have a happy path here. I'm going to address this.

example-gto/code/.github/workflows/gto-act-on-tags.yml
```diff
@@ -0,0 +1,46 @@
...
+        git fetch --all --prune --tags
+        git pull --all
+        set -e
+    - uses: actions/setup-python@v2
+    - name: set PYSHA
+      run: echo "PYSHA=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
+    - uses: actions/cache@v1
+      with:
+        path: ~/.cache/pre-commit
+        key: pre-commit|${{ env.PYSHA }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
shcheklein, line 29: do we even have this file?

example-gto/code/.github/workflows/gto-act-on-tags.yml
```diff
@@ -0,0 +1,46 @@
+name: Act on artifact registrations and promotions
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  act:
+    name: Check
```
shcheklein, line 9: what does it check (I can't understand the logic of this workflow tbh) - seems we have some things that are not needed, but I'm not sure

example-gto/code/README.md
```diff
@@ -0,0 +1,33 @@
+# GTO Get Started
+
+This is an auto-generated repository for use in
+[GTO Get Started](https://github.com/iterative/gto/blob/main/README.md).
```
shcheklein, line 4: can be a just link to the repo itself I think

example-gto/code/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+```console
+$ virtualenv -p python3 .env
+$ source .env/bin/activate
+$ pip install -r requirements.txt
+```
+
+## To see GTO in action
+
+To understand how this repo is generated, take a look at the
```
shcheklein, line 31: should we redirect to the Get Started? (high level vs this generate.sh that has a lot of things that are not that relevant)

example-gto/generate.sh
shcheklein: sorry, I meant to ask - is there a specific reason to have it here? (it can be error prone if you see that venv is not activated)

example-gto/generate.sh
```diff
@@ -0,0 +1,184 @@
...
+unset GIT_AUTHOR_DATE
+unset GIT_COMMITTER_DATE
+unset GIT_AUTHOR_NAME
+unset GIT_AUTHOR_EMAIL
+unset GIT_COMMITTER_NAME
+unset GIT_COMMITTER_EMAIL
+
+cat <<EOF
+
+The Git repo generated by this script is intended to be published on
```
shcheklein, line 163: Should we update this - looks like the way to push it is to use PUSH=true or something like that?

example-gto/code/.github/workflows/gto-act-on-tags.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: set PYSHA
+      run: echo "PYSHA=$(python -VV | sha256sum | cut -d' ' -f1)" >> $GITHUB_ENV
+    - uses: actions/cache@v1
+      with:
+        path: ~/.cache/pre-commit
+        key: pre-commit|${{ env.PYSHA }}|${{ hashFiles('.pre-commit-config.yaml') }}
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip setuptools wheel
+        pip install -r requirements.txt
```
shcheklein, line 33: looks like the requirements file is missing here?

example-gto/generate.sh
shcheklein: not wrong, not a usual way (usually you expect just regular deps) and I would expect Pipfile.lock for this. I haven't seen this approach for the requirements for a while.

anyways, in this case it's also bad because you have to generate the new project anytime you update GTO, otherwise it's get stuck with the previous gto version

example-gto/generate.sh
aguschin: Don't know TBH, just copied this over from mlem PR. Let's delete it :)

example-gto/generate.sh
aguschin: Ok, fixing this.

example-gto/code/.github/workflows/gto-act-on-tags.yml
aguschin: Thanks, removed.

example-gto/code/.github/workflows/gto-act-on-tags.yml
aguschin: That was a wrong name, renamed it. Removed extra stuff that's not needed.

example-gto/code/README.md
aguschin: Done

example-gto/code/README.md
aguschin: Good idea, added the link.

example-gto/generate.sh
aguschin: Yes, thanks! Did that.

example-gto/code/.github/workflows/gto-act-on-tags.yml
aguschin: No, it's there. Earlier is was generated by `pip freeze` and now it's just exists in `code/` folder.

example-gto/code/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+```console
+$ virtualenv -p python3 .env
+$ source .env/bin/activate
+$ pip install -r requirements.txt
+```
+
+## To see GTO in action
+
+Go ahead to the [GTO Getting Started](https://github.com/iterative/gto#readme) to try out the tool.
```
shcheklein, line 31: formatting, please run linters to keep it under 80 symbols

example-gto/generate.sh
```diff
@@ -0,0 +1,174 @@
...
+appropriate write permissions.
+
+To run the generator in the test mode, just use "bash generate.sh"
+
+To push it to GitHub, use "bash generate.sh --push"
+This will do it step by step waiting for CI to have consistent results.
+
+To cd to the generated repo, run "cd build/example-gto"
+
+You may remove the generated repo with "rm -fR build"
```
shcheklein, line 173: minor: `rm -fR build/example-gto`

shcheklein (approved):
minor comments, looks good to me otherwise, thanks @aguschin !

aguschin:
@shcheklein, please see the new GH actions example. It doesn't do real publish/deploy, but it should look more confident now.

https://github.com/iterative/example-gto/actions
image

PR #112, opened 2022-04-22
get-started: fix CML workflow

shcheklein:
Use actions instead of containers
Use fetch-depth 2 to fix action running on master


0x2b3bfa0 (approved):
Looks good to me 🎉

PR #117, opened 2022-05-20
Add flexible plots

dberenbaum:
Need to wait for https://github.com/iterative/dvc/pull/7477. May also need/want to adjust syntax depending on final state of that core PR.

Train metrics/plots look bad and show that the model is overfitting. Adjusting `train.min_samples_split` to fractional float value (min_samples_split: 0.01 to split until <0.01 of data is in the leaf) make the results more typical.

PR #118, opened 2022-05-22
get-started: move importance into plots

shcheklein:
Based on @maxagin 's feedback , let's make it symmetrical and explicit that `importance.png` is also a plot.

### DIFF
--- a/example-get-started/code/README.md
+++ b/example-get-started/code/README.md
@@ -139,27 +139,27 @@ data, models, and plots tracked by DVC, the workspace should look like this:
 $ tree
 .
 ├── README.md
-├── data                  # <-- Directory with raw and intermediate data
-│   ├── data.xml          # <-- Initial XML StackOverflow dataset (raw data)
-│   ├── data.xml.dvc      # <-- .dvc file - a placeholder/pointer to raw data
-│   ├── features          # <-- Extracted feature matrices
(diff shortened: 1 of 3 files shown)


dberenbaum:
This may be more complicated than @maxagin realized 😅. It looks good in the example, but it may set a bad pattern for users that might not work generally. I might be missing some context, but let me explain my understanding of the situation.

DVCLive logs different types of data in up to 3 different folders: scalars, plots, and images (https://dvc.org/doc/dvclive/get-started#outputs). More confusingly, all of these could be plots.

I don't think putting image files into the plots folder will make sense when the other folders are also populated (it might make sense under images if that folder existed). I would find it confusing to have the same folder contain my json and image files. 

It might be better to focus on how those folders should be named in DVCLive, as mentioned in https://github.com/iterative/dvclive/issues/246. Considering that all of these folders can include plots, it's probably best not to reserve the plots name for any specific DVCLive folder.

dberenbaum:
If we want to do something like https://github.com/iterative/example-repos-dev/pull/117, then the other plots will all have separate train and test folders while `importance.png` will not.

shcheklein:
@dberenbaum okay, sounds good and good points, https://github.com/iterative/dvclive/issues/246 should be a high priority then, since it's a public convention, right? it's not good that we'll be breaking it so soon.

We can close this for now.

PR #119, opened 2022-05-23
get-started: use fractional min_split

dberenbaum:
Using a fractional value for `min_split` (see https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html) reduces overfitting and scales better to different sample sizes.

train.min_split=2

visualization(6)

train.min_split=0.01

visualization(7)


shcheklein (approved):

shcheklein:
@dberenbaum thanks, I'll regenerate the project with this change.

PR #122, opened 2022-06-03
Fixture for Model Registry with GTO+MLEM+DVC

aguschin:
This is a fixture that generates bunch of models tracked with GTO, MLEM, DVC or all of those. Not sure it's eventually should be here, but since I generate it using the same approach, it could be the right place.

This is the current result of running it: http://github.com/aguschin/fixture-model-registry

You'll need to use was and azure keys to run it (`generate.sh` will warn you about this).


fixture-model-registry/code/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "experiments-get-started",
```
shcheklein, line 2: should be different?

should we install DVC extension also?

why do we need codespaces for this btw?

fixture-model-registry/code/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,7 @@
...
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
+  "postCreateCommand": "pip3 install --user mlem"
```
shcheklein, line 6: looks like mlem is part of the requirements already, should we just install all of them (including GTO)?

fixture-model-registry/code/README.md
```diff
@@ -0,0 +1,39 @@
+# MLEM Get Started
```
shcheklein, line 1: it's not get started though?

fixture-model-registry/code/README.md
```diff
@@ -0,0 +1,39 @@
...
+[example-repos-dev](https://github.com/iterative/example-repos-dev).
+
+## Installation
+
+Python 3.7+ is required to run code from this repo.
+
+To be able to commit and push your results please fork and then clone this repo
+
+```console
+$ gh repo fork https://github.com/iterative/example-mlem-get-started
```
shcheklein, line 20: rename? (and review this README please)

fixture-model-registry/generate.sh
```diff
@@ -0,0 +1,203 @@
...
+
+echo "Create new models"
+dvc init
+mlem init
+tick
+git add .mlem .dvc
+git commit -m "Initialize MLEM and DVC project"
+
+
+python train.py
```
shcheklein, line 81: Q: do we need actual code / train or we can imitate it for the fixture?

shcheklein:
@aguschin close as stale?

aguschin:
@shcheklein, let's see if the repo produced by this script is used for Studio. If not, we can close it as stale.

@amritghimire, do you use your fork of https://github.com/aguschin/fixture-model-registry to perform tests for Studio?

If the answer would be "yes", this is going to be useful in future, when we make some breaking changes to GTO and will need to re-generate it (e.g. if we move `artifacts.yaml` to `dvc.yaml`).

shcheklein:
@aguschin Q: do we need this still?

aguschin:
@shcheklein I just re-generated fixture repo for Studio a week ago with this, and it saved us a lot of time (I didn't need to reproduce the repo manually, @amritghimire needed less fixes in tests due to discrepancies in regenerated repo). We may need this in future, I assume. If you don't want this to be here, I can fork a repo and keep this branch on my personal account, while closing PR and removing the branch here. Just let me know if this is preferred.

shcheklein:
@aguschin no, not at all. I'm all in to merge this if it saves time and it's valuable! I was just asking if we want to make a decision and if this is needed or not. Since it benefits us, let's merge it, wdyt? Do we need to do any additional fixes for this?

fixture-model-registry/code/README.md
aguschin: Yep, updated!

fixture-model-registry/code/.devcontainer/devcontainer.json
aguschin: done

fixture-model-registry/generate.sh
aguschin: Let's keep it - it create some diversity in trained models (we have some "dummy ML models" saved already). Plus the fact it's a real MLEM model can be used in tests I think.

aguschin:
Great @shcheklein! I've updated the branch fixing the feedback you left initially :) Please review and let's merge.

shcheklein (approved):

shcheklein:
Thanks @aguschin !

PR #125, opened 2022-07-04
Update to use `dvc.api.params_show`

daavoo:


shcheklein (changes requested):
I would not do this for this project. DVC has to be installed as lib then and currently it's not (it's not part of the requirements file). Codespaces setup installs it from deb.

Overall, can this kind of API be a separate package?

dberenbaum:
Would that package need DVC as a dependency to collect the parameters for each stage?

daavoo:
> Overall, can this kind of API be a separate package?

I think it won't change much as it would require `dvc`.

PR #126, opened 2022-07-18
Shorten params to use keys only

dberenbaum:
Small change from `dvc run -p prepare.seed,prepare.split` -> `dvc run -p prepare`


dberenbaum:
Requires accompanying text changes in the docs.

shcheklein:
I think we should keep more explicit, tbh. Or at least, keep it explicit in the first place and make it short in the second and the last. I think the logic behind this was - it's simpler for readers to see the names and recognize the dependency. WDYT?

dberenbaum:
Okay, I thought you might say that 😄 . I don't have a strong opinion and can see the benefits of both. Closing for now.

Issue #129, opened 2022-08-18
Reverse order of `deploy.sh` and `generate.sh`


dberenbaum:
I always find it confusing when developing on this repo that I have to deploy the code I'm still developing before generating the repo locally. I end up having to modify the scripts so I don't accidentally deploy the untested code but also can still use it to generate a local copy of the repo for development. Would it make sense to modify the scripts so that `generate.sh` uses whatever code is locally available and `deploy.sh` is only used as part of the final deployment process (maybe as part of a GH action when merged)?

shcheklein:
It makes sense, I would though probably make `generate.sh prod` or something that still downloads from S3 (so that we can be 100% that we do exactly same steps as users would do).

PR #130, opened 2022-08-18
example-get-started: make dev/prod generate.sh and source deploy.sh

dberenbaum:
@shcheklein Here's a proposed attempt to address #129. Let me know what you think.


shcheklein (approved):

shcheklein:
Thanks @dberenbaum ! I think we can merge, try, adjust a bit if needed - looks good to me

PR #136, opened 2022-09-18
get-started: add shap plot

dberenbaum:
Started on a [SHAP plot](https://christophm.github.io/interpretable-ml-book/shap.html#shap-summary-plot). This is a popular method for feature importance that's more robust than the ones built in to sklearn. It provides an example of a plot that needs to be saved as an image since it would be hard to support as a vega plot.

Example output:

image

One downside is that it's expensive, so it might make sense to sample the data here, but I didn't bother for now.


dberenbaum:
Closing as stale

PR #140, opened 2022-09-30
Update use of cml cli

tasdomas:
The cli for cml has been updated, it's best if our examples use the new command structure.

### DIFF
--- a/example-get-started/code/.github/workflows/cml.yaml
+++ b/example-get-started/code/.github/workflows/cml.yaml
@@ -24,7 +24,7 @@ jobs:
           cat <<EOF > report.md
           # CML Report
           ## Plots
-          ![Prevision & Recall](./prc.svg)
+          ![Precision & Recall](./prc.svg)
           ![Confusion Matrix](./confusion.svg)
 
           ## Metrics and Params
--- a/example-get-started/code/.github/workflows/cml.yaml
+++ b/example-get-started/code/.github/workflows/cml.yaml
@@ -20,16 +20,6 @@ jobs:
             PREVIOUS_REF=main
           fi
 
-          cat <<EOF > report.md
-          # CML Report
-          ## Plots
-          ![Precision & Recall](./prc.svg)
-          ![Confusion Matrix](./confusion.svg)
-
-          ## Metrics and Params
-          ### ${PREVIOUS_REF} → workspace
-          EOF
-
           dvc plots diff $PREVIOUS_REF workspace \
             --show-vega --targets evaluation/plots/precision_recall.json > vega.json
           vl2svg vega.json prc.svg
@@ -38,6 +28,17 @@ jobs:
             --show-vega --targets evaluation/plots/confusion_matrix.json > vega.json
           vl2svg vega.json confusion.svg
 
-          dvc exp diff $PREVIOUS_REF --show-md >> report.md
+          dvc_report=$(dvc exp diff $PREVIOUS_REF --show-md)
+
+          cat <<EOF > report.md
+          # CML Report
+          ## Plots
+          ![Precision & Recall](./prc.svg)
+          ![Confusion Matrix](./confusion.svg)
+
+          ## Metrics and Params
+          ### ${PREVIOUS_REF} → workspace
+          ${dvc_report}
+          EOF
 
           cml comment create --publish --pr=false report.md
--- a/example-get-started/code/.github/workflows/cml.yaml
+++ b/example-get-started/code/.github/workflows/cml.yaml
@@ -8,7 +8,6 @@ jobs:
       - uses: iterative/setup-dvc@v1
       - uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 2
       - name: Generate metrics report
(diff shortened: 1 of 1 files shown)


example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -8,6 +8,7 @@ jobs:
       - uses: iterative/setup-dvc@v1
       - uses: actions/checkout@v3
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
shcheklein, line 11: @tasdomas Q: why do we need this?

tasdomas: For `cml comment create --pr` (comments on PRs) to work, cml needs to be able to determine the PR id based on the HEAD commit. However github runs workflows on a {master..PR} merge commit, which cml currently cannot attribute to the actual PR. This is a temporary workaround until we implement proper PR lookup in cml.

shcheklein: got it. In this repo we were also publishing comments to the main branch commits:

https://github.com/iterative/example-repos-dev/pull/140/files#diff-74446cb71a98e9596cef50d831916c75ed4f0c65a5083c75dca8de15cdd5dd50R18-R22

what will happen to those with this new change (and with `--pr`)?

I mean reports like this https://github.com/iterative/example-get-started/commit/a470fae5c0ccf71641e10d99e69e97534c3d811e#commitcomment-84532489

tasdomas: There's an issue with commit PRs, that affects CML - in github they [no longer appear in the PR timeline](https://github.blog/changelog/2022-08-04-commit-comments-no-longer-appear-in-the-pull-request-timeline/) which may make adding commit comments in our example repos fall short in showcasing CML functionality.

@casperdcl  - what do you think?

casperdcl: > In this repo we were also publishing comments to the main branch commits

afaik `--pr` should fallback to non-`--pr` in such cases.

Alternatively should we change this workflow from `on: push` to `on: pull_request` (or both `push` & `pull_request`)? To me the `main` monitoring use case is different from the PR use case, so maybe even separate workflows?

shcheklein: yep, would be a bit annoying to duplicate the workflow file in this repo in this simple case though

tasdomas: I've revisited this and if all we want is commit tracking, then we don't need this change. Sorry, I was a tad over-zealous.

example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -20,20 +20,24 @@ jobs:
...
+          cat <<EOF > report.md
+          # CML Report
+          ## Plots
+          ![Precision & Recall](./prc.svg)
+          ![Confusion Matrix](./confusion.svg)
+
+          ## Metrics and Params
+          ### ${PREVIOUS_REF} → workspace
+          EOF
+
```
casperdcl, line 32:
```suggestion
```

example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -20,20 +20,24 @@ jobs:
...
-          cml publish prc.svg --title "Precision & Recall" --md >> report.md
 
           dvc plots diff $PREVIOUS_REF workspace \
             --show-vega --targets evaluation/plots/confusion_matrix.json > vega.json
           vl2svg vega.json confusion.svg
-          cml publish confusion.svg --title "Confusion Matrix" --md >> report.md
 
-          echo "## Metrics and Params" >> report.md
-          echo "### $PREVIOUS_REF → workspace" >> report.md
           dvc exp diff $PREVIOUS_REF --show-md >> report.md
```
casperdcl, line 41:
```suggestion
          cat <<EOF > report.md
          # CML Report
          ## Plots
          ![Precision & Recall](./prc.svg)
          ![Confusion Matrix](./confusion.svg)

          ## Metrics and Params
          ### ${PREVIOUS_REF} → workspace
          $(dvc exp diff $PREVIOUS_REF --show-md)
          EOF
```

casperdcl (approved):
minor suggestion to tidy code a bit

shcheklein:
@iterative/cml what is the status of this?

tasdomas:
@shcheklein - once this is landed, how do we release the changes? I suspect I will not have access to all the necessary cloud resources

example-get-started/code/.github/workflows/cml.yaml
tasdomas: executing a command in the middle of a heredoc gives me the shivers

tasdomas: Submitted an alternative - ptal

tasdomas:
@shcheklein the functionality will remain unchanged - the comments will be created on pushed commits

shcheklein:
thanks @tasdomas @casperdcl ... I'll update the repo (I'm stuck a bit since there is something happening with DVC)

PR #142, opened 2022-11-04
example-get-started: dvclive 1.0 updates

dberenbaum:
Wait for dvclive 1.0 release

### DIFF
--- a/example-get-started/generate.sh
+++ b/example-get-started/generate.sh
@@ -142,17 +142,19 @@ dvc push
 
 dvc run -n evaluate \
   -d src/evaluate.py -d model.pkl -d data/features \
-  -M evaluation.json \
-  --plots-no-cache evaluation/plots/precision_recall.json \
-  --plots-no-cache evaluation/plots/roc.json \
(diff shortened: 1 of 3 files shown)


example-get-started/generate.sh
```diff
@@ -194,7 +196,7 @@ dvc exp run --queue --set-param train.min_split=8 --set-param train.n_est=100
 dvc exp run --queue --set-param train.min_split=64 --set-param train.n_est=100
 dvc exp run --run-all -j 2
 # Apply best experiment
-dvc exp apply $(dvc exp show --no-pager --sort-by avg_prec | tail -n 2 | head -n 1 | grep -o 'exp-\w*')
+dvc exp apply $(dvc exp show --sort-by avg_prec --csv | tail -n 1 | cut -d , -f 1)
```
dberenbaum, line 199: Unrelated but this was causing problems with a narrow terminal where exps were shown on multiple lines, so this should be more stable.

daavoo (approved):

PR #144, opened 2022-11-05
example-dvc-experiments: remove upper bound on tensorflow

dberenbaum:
Seems like the requirements are a bit overly restrictive here and the pinned versions are no longer available. I think it's enough to have a lower bound (if there's a tensorflow 3.0, we can revisit).


dberenbaum:
It's already pushed to https://github.com/iterative/example-dvc-experiments

shcheklein (approved):

PR #147, opened 2022-11-06
example-get-started: fix plot paths

0x2b3bfa0:
`dvc plots diff` target paths pointed to non-existent locations and caused `vl2svg` to fail.

### DIFF
--- a/example-get-started/code/.github/workflows/cml.yaml
+++ b/example-get-started/code/.github/workflows/cml.yaml
@@ -22,11 +22,11 @@ jobs:
           fi
 
           dvc plots diff $PREVIOUS_REF workspace \
-            --show-vega --targets evaluation/plots/precision_recall.json > vega.json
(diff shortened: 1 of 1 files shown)


shcheklein:
Thanks @0x2b3bfa0 ! We should probably make sure that it fails if one of the commands fails (and make sure that plots fail in this case, probably CML report also should fail?)

shcheklein (approved):

0x2b3bfa0:
The `vl2svg` command exits with code `0` no matter what, so there is no elegant solution on that front. As well, CML doesn't fail when a file doesn't exist, because it can also be remote URL. 😅 

Maybe DVC should exit with non-zero status if the `--target` doesn't exist?

0x2b3bfa0:
Alternatively, there are “solutions” like this: 🙃 

```
test -s input.json
vl2svg $_ output.svg
```

PR #154, opened 2022-12-02
example-get-started: `dvc run` -> `dvc stage add`

### LINKED ISSUE #153: `example-get-started`: convert `dvc run` to `dvc stage add` in `generate.sh`
We have moved to using `dvc stage add` instead of `dvc run` throughout the docs, but `generate.sh` still uses `dvc run`. This actually causes some problems because of differences in behavior like https://github.com/iterative/dvc/issues/5802, so users may be running into errors while following the tutorial. We either need to ensure the workflow works with `stage add` or revert to `run`.

dberenbaum:
Fixes #153


shcheklein (approved):

shcheklein (approved):

dberenbaum:
Don't think we need to redeploy for this.

PR #160, opened 2023-01-19
example-get-started: update plots ordering

dberenbaum:
See https://github.com/iterative/dvc.org/pull/4207#discussion_r1068268762. tldr: This updates better matches the text in https://dvc.org/doc/start/data-management/metrics-parameters-plots#visualizing-plots.


shcheklein (approved):

dberenbaum:
@shcheklein Do you think it's worth redeploying for this? I think there are still some manual steps involved, right?

shcheklein:
@dberenbaum no, I think we are fine.  We can redeploy later when we have more updates. Your call though.

Issue #179, opened 2023-02-03
experiments: some DVC artifacts are missing when running `dvc pull` 


alex000kim:
- Launched Codespaces on `main` in https://github.com/iterative/example-get-started-experiments/
- Ran `dvc pull`
- See this error
```
@alex000kim ➜ /workspaces/example-get-started-experiments (main) $ dvc pull
WARNING: Some of the cache files do not exist neither locally nor on remote. Missing cache files:                                                                                                         
name: models/model.pkl, md5: 876a7dc3312f64f66d245242fdc03b41                                                                                                                                             
1 file failed                                                                                                                                                                                             
ERROR: failed to pull data from the cloud - Checkout failed for following targets:                                                                                                                        
/workspaces/example-get-started-experiments/models/model.pkl                                                                                                                                              
Is your cache up to date?                                                                                                                                                                                 
<https://error.dvc.org/missing-files>
```

alex000kim:
I think it has something to do with this message when running `bash generate.sh`:

```bash
...
+ git commit -am 'Run experiments tuning architecture. Apply best one'
[tune-architecture c842192] Run experiments tuning architecture. Apply best one
 22 files changed, 45 insertions(+), 45 deletions(-)
+ git checkout main
Switched to branch 'main'
+ dvc push -A
WARNING: Output 'data/test_data'(stage: '../../../../../dvc.yaml:data_split') is missing version info. Cache for it will not be collected. Use `dvc repro` to get your pipeline up to date.
WARNING: Output 'data/train_data'(stage: '../../../../../dvc.yaml:data_split') is missing version info. Cache for it will not be collected. Use `dvc repro` to get your pipeline up to date.
WARNING: Output 'models/model.pkl'(stage: '../../../../../dvc.yaml:train') is missing version info. Cache for it will not be collected. Use `dvc repro` to get your pipeline up to date.
ERROR: unexpected error - Forbidden: An error occurred (403) when calling the HeadObject operation: Forbidden                                                                                                             

Having any troubles? Hit us up at https://dvc.org/support, we are always happy to help!
```

alex000kim:
even after running `dvc repro; dvc push -A` I am still getting the same warnings
```
(.venv) @alex000kim ➜ .../example-repos-dev/example-get-started-experiments/build/example-get-started-experiments (main) $ dvc repro -f; dvc push -A
Verifying data sources in stage: 'data/pool_data.dvc'                 
                                                                                                                                                                                                                          
[...]

PR #182, opened 2023-02-03
example-get-started-experiments suggested update re: Studio Live Metrics

jendefig:
Suggestions, given what I'm finding when I try it out.

### DIFF
--- a/example-get-started-experiments/code/README.md
+++ b/example-get-started-experiments/code/README.md
@@ -50,7 +50,6 @@ You can run [`dvc pull`](https://man.dvc.org/pull) to download the data:
 ```console
 $ dvc pull
 ```
-If you get an error when you `dvc pull`, don't worry you will still be able to run your experiments, this is because (of DVC shared cache? guessinghttps://dvc.org/doc/user-guide/how-to/share-a-dvc-cache#configure-the-shared-cache)
 
 ## Running in your environment
 
--- a/example-get-started-experiments/code/README.md
+++ b/example-get-started-experiments/code/README.md
@@ -50,8 +50,11 @@ You can run [`dvc pull`](https://man.dvc.org/pull) to download the data:
 ```console
 $ dvc pull
 ```
+If you get an error when you `dvc pull`, don't worry you will still be able to run your experiments, this is because (of DVC shared cache? guessinghttps://dvc.org/doc/user-guide/how-to/share-a-dvc-cache#configure-the-shared-cache)
 
 ## Running in your environment
+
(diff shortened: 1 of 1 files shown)


example-get-started-experiments/code/README.md
```diff
@@ -50,9 +50,12 @@ You can run [`dvc pull`](https://man.dvc.org/pull) to download the data:
 ```console
 $ dvc pull
 ```
+If you get an error when you `dvc pull`, don't worry you will still be able to run your experiments, this is because (of DVC shared cache? guessinghttps://dvc.org/doc/user-guide/how-to/share-a-dvc-cache#configure-the-shared-cache)
```
shcheklein, line 53: I don't think we need this. There should be no errors. It's a bug if that is not the case.

example-get-started-experiments/code/README.md
```diff
@@ -50,9 +50,12 @@ You can run [`dvc pull`](https://man.dvc.org/pull) to download the data:
...
 
 ## Running in your environment
 
+If you plan to use Iterative Studio to view real-time, live metrics, please refer to [these instructions](https://dvc.org/doc/studio/user-guide/projects-and-experiments/live-metrics-and-plots) first.
```
shcheklein, line 57: Let's make it after the `dvc exp run` code block. And make it something like:

To visualize metrics, experiments in real-time you could use Studio (make it a link) ... or VS Code extension ... or CLI `dvc exp show`

shcheklein: also, please keep it <80 as we do in this file

example-get-started-experiments/code/README.md
daavoo, line 53:
```suggestion
```

jendefig: @alex000kim Can explain more.  We intend to use this project for the DVC Live Metircs release and when you follow the instructions of the README, this error occurs.  So do we need to hold this until this is not the case?

alex000kim: There used to be an issue with `dvc pull`, but it fixed now: https://github.com/iterative/example-repos-dev/issues/179

jendefig: Yay!  🙏

example-get-started-experiments/code/README.md
jendefig: Ok.  Can someone tell me how I correct a patch because I have no idea?  Or do I just close this and start over with another patch?

jendefig: Thanks @dberenbaum!  Everything changed so much since then, I don't even know if this is relevant anymore.

Issue #183, opened 2023-02-03
experiments: check if changes needed for compatibility with dvclive 2.0


dberenbaum:
Two potential issues come to mind:
1. Plots config in dvc.yaml may need to be updated per https://github.com/iterative/dvclive/pull/444.
2. May need to toggle off `Live(dvcyaml=True)` to avoid duplicate plots.

PR #185, opened 2023-02-06
Experiments followups

daavoo:
Closes https://github.com/iterative/example-repos-dev/issues/183

### DIFF
--- a/example-get-started-experiments/generate.sh
+++ b/example-get-started-experiments/generate.sh
@@ -117,6 +117,7 @@ dvc stage add -n evaluate \
   --plots-no-cache results/evaluate/plots \
   python src/evaluate.py
 
+rm results/train/dvc.yaml
 git rm --cached results/train/dvc.yaml
 git add .
 tick
--- a/example-get-started-experiments/generate.sh
+++ b/example-get-started-experiments/generate.sh
@@ -15,7 +15,7 @@ if [ -d "$REPO_PATH" ]; then
   exit 1
 fi
 
-TOTAL_TAGS=3
+TOTAL_TAGS=8
 STEP_TIME=100000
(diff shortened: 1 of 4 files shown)


example-get-started-experiments/generate.sh
```diff
@@ -116,6 +116,9 @@ dvc stage add -n evaluate \
   -M results/evaluate/metrics.json \
   --plots-no-cache results/evaluate/plots \
   python src/evaluate.py
+
+rm results/train/dvc.yaml
+git rm --cached results/train/dvc.yaml
```
daavoo, line 121: @dberenbaum I don't know what would be the best way to handle the transition to pipelines. 
Should we do this removal or instead don't include the `metrics `and `plots` sections in the root `dvc.yaml`?

dberenbaum: @daavoo I would prefer the latter. It seems easier to keep `Live(dvcyaml=True)` and change `dvc stage add -M results/train/metrics.json --plots-no-cache results/train/plots` to `dvc stage add -O results/train` unless I'm missing something.

I would do the same for the evaluate stage (also, why don't we evaluate the test data in the notebook?).

daavoo: > (also, why don't we evaluate the test data in the notebook?)

We do. But when converting to pipeline it gets separated into a different stage

dberenbaum: > change `dvc stage add -M results/train/metrics.json --plots-no-cache results/train/plots` to `dvc stage add -O results/train` unless I'm missing something

This needs to be `dvc stage add -O results/train/metrics.json -O results/train/plots` for now at least. See https://github.com/iterative/dvclive/issues/456. I still think that's better since we only need to introduce dependencies and outputs when we get to pipelines instead of special output types like metrics and plots.

PR #191, opened 2023-03-23
experiments: use Live(dvcyaml=True)

### LINKED ISSUE #190: experiments: use dvcyaml=true
Is there a reason we use `Live(dvcyaml=False)` in the scripts @alex000kim @daavoo? We don't do any custom plotting, so it seems like it would be easier to leave them on and avoid needing to learn about metrics and plots outputs in the pipelines.

dberenbaum:
Fixes #190. Unfortunately I haven't been able to regenerate the repo since I don't have GPU access, so I'm leaving this PR in draft mode.

### DIFF
--- a/example-get-started-experiments/generate.sh
+++ b/example-get-started-experiments/generate.sh
@@ -106,15 +106,11 @@ dvc stage add -n train \
   -p base,train \
   -d src/train.py -d data/train_data \
   -o models/model.pkl \
-  -O results/train/metrics.json \
-  -O results/train/plots \
   python src/train.py
 
 dvc stage add -n evaluate \
   -p base,evaluate \
   -d src/evaluate.py -d models/model.pkl -d data/test_data \
-  -O results/evaluate/metrics.json \
-  -O results/evaluate/plots \
   python src/evaluate.py
 
 rm results/train/dvc.yaml
--- a/example-get-started-experiments/generate.sh
+++ b/example-get-started-experiments/generate.sh
@@ -106,15 +106,15 @@ dvc stage add -n train \
   -p base,train \
   -d src/train.py -d data/train_data \
   -o models/model.pkl \
-  -M results/train/metrics.json \
-  --plots-no-cache results/train/plots \
+  -O results/train/metrics.json \
+  -O results/train/plots \
   python src/train.py
 
 dvc stage add -n evaluate \
   -p base,evaluate \
   -d src/evaluate.py -d models/model.pkl -d data/test_data \
-  -M results/evaluate/metrics.json \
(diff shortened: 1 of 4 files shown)


example-get-started-experiments/generate.sh
```diff
@@ -106,15 +106,15 @@ dvc stage add -n train \
   -p base,train \
   -d src/train.py -d data/train_data \
   -o models/model.pkl \
-  -M results/train/metrics.json \
-  --plots-no-cache results/train/plots \
+  -O results/train/metrics.json \
+  -O results/train/plots \
```
dberenbaum, line 110: Also happy to drop these completely, which would simplify a lot of output, but want to coordinate this between here, https://github.com/iterative/dvc.org/pull/4420, and https://github.com/iterative/dvclive/pull/497.

daavoo: I vote in favor of going all-in with  the dropping approach across all places. 

We still need to find a place/way to explain how the UIs are finding the stuff to be displayed in this scenario, though

dberenbaum: Updated to drop them.

Where are you thinking to explain them? We have https://dvc.org/doc/dvclive/live/make_dvcyaml, and we could maybe reference it in https://dvc.org/doc/dvclive/how-it-works#track-the-results. Do you think we need it anywhere else?

dberenbaum:
@daavoo @alex000kim Could one of you try this out on the GPU runner when you have a chance so we could merge this PR?

daavoo:
> @daavoo @alex000kim Could one of you try this out on the GPU runner when you have a chance so we could merge this PR?

on it

daavoo:
Ok, it appears that something might be broken on DVC side when there are 2 dvc.yaml files with top-level stuff. Only one is being collected. 😫 

Taking a look

daavoo:
> Ok, it appears that something might be broken on DVC side when there are 2 dvc.yaml files with top-level stuff. Only one is being collected. 😫
> 
> Taking a look

It was 

> Ok, it appears that something might be broken on DVC side when there are 2 dvc.yaml files with top-level stuff. Only one is being collected. 😫
> 
> Taking a look

I think it was https://github.com/iterative/example-repos-dev/pull/191#issuecomment-1492252216

daavoo:
The deployed version already contains these changes (https://github.com/iterative/example-get-started-experiments/blob/main/dvc.yaml) but I was trying to also push the experiments with `exp push` and failing.
Probably no reason to don't merge this P.R.

daavoo (approved):

dberenbaum:
Thanks @daavoo!

PR #192, opened 2023-03-31
`example-gto`: remove `annotate` and move artifacts to `dvc.yaml`

aguschin:
- the repo can be re-generated already, although CI will fail
- we need to update GTO action to v2 to make it work in CI
- test repo for Studio BE should have a monorepo. Given `example-gto`, one can ~generate monorepo with simply
```sh
$ cp -r * example-gto example-gto/subrepo
$ cd example-gto
```
Now you'll have a very simple monorepo 🙌🏻 

cc @jellebouwman and @amritghimire re https://github.com/iterative/studio/issues/5504 - I'm not sure `demo-bank-customer-churn` is a monorepo, so if you need one, we can do this ^

### DIFF
--- a/example-gto/code/mlem/.github/workflows/deploy-model-with-mlem.yml
+++ b/example-gto/code/mlem/.github/workflows/deploy-model-with-mlem.yml
@@ -16,13 +16,15 @@ jobs:
     - uses: actions/checkout@v3
     - name: "GTO: figure out what was registered/promoted and show the Registry state"
       id: gto
-      uses: iterative/gto-action@v1
+      uses: iterative/gto-action@v2
+      with:
+        download: true
     # we define the Job outputs here to let the next Job use them
     outputs:
       name: ${{ steps.gto.outputs.name }}
       stage: ${{ steps.gto.outputs.stage }}
       event: ${{ steps.gto.outputs.event }}
-      path: ${{ steps.gto.outputs.path }}
+      path: ${{ steps.gto.outputs.path }}  # the path the model was `dvc get` to
   deploy-model:
     name: Deploy a MLEM model (act on assigning a new stage)
     needs: parse-git-tag
--- a/example-gto/code/mlem/.github/workflows/deploy-model-with-mlem.yml
+++ b/example-gto/code/mlem/.github/workflows/deploy-model-with-mlem.yml
@@ -24,7 +24,7 @@ jobs:
       name: ${{ steps.gto.outputs.name }}
       stage: ${{ steps.gto.outputs.stage }}
       event: ${{ steps.gto.outputs.event }}
-      path: ${{ steps.gto.outputs.path }}  # the path the model was `dvc get` to
+      path: ${{ steps.gto.outputs.path }}
   deploy-model:
     name: Deploy a MLEM model (act on assigning a new stage)
     needs: parse-git-tag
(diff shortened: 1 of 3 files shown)


amritghimire:
> the repo can be re-generated already, although CI will fail

For studio, this wont be an issue since we use the frozen repo for tests. We can take care of the test cases in upgrade of GTO. I agree we can add mono repo for the test cases. But I think even with upgrade, the repo with the old artifacts.yaml and the information in dvc.yaml should both work. Or at the very least, when doing some action like registration and so on, it should migrate the data accordingly. WDYT?

jellebouwman:
>  I'm not sure demo-bank-customer-churn is a monorepo, so if you need one, we can do this ^

I don't think it is, it's adding different types of models on different branches, but everything is added to the root of the repository!

aguschin:
> But I think even with upgrade, the repo with the old artifacts.yaml and the information in dvc.yaml should both work.

To smoothen migration, we can implement if-else, supporting both old-format GTO repo and new format. Something like:
```
artifacts = dvc.repo.Repo(".").artifacts
if not artifacts:
  artifacts = gto.api.show(".")
```

> Or at the very least, when doing some action like registration and so on, it should migrate the data accordingly. WDYT?

GTO API will not change regarding registrations/assignments, but annotation will be removed, so we'll need to use internal DVC API for that, or update `dvc.yaml` manually.

So, what should be done in Studio (I'm surely missing something though):
1. Updating DVC version used (we're yet to release it in DVC though)
2. Supporting old (`artifacts.yaml`, GTO API) / new(`dvc.yaml`, DVC API) formats for reading (this `try..except` ^)
3. Supporting new format for annotation, removing old format for annotation
4. Adding new test repo (re-generated `example-gto`)
5. Showing a warning for old format users, telling them they should migrate

@amritghimire - I'm happy to start drafting this in BE, let's have a call so you can point me towards right modules/scripts? Thanks!

aguschin:
@omesser can you please approve and merge? Thanks!

example-gto/code/mlem/.github/workflows/deploy-model-with-mlem.yml
```diff
@@ -16,13 +16,15 @@ jobs:
...
+      uses: iterative/gto-action@v2
+      with:
+        download: true
     # we define the Job outputs here to let the next Job use them
     outputs:
       name: ${{ steps.gto.outputs.name }}
       stage: ${{ steps.gto.outputs.stage }}
       event: ${{ steps.gto.outputs.event }}
-      path: ${{ steps.gto.outputs.path }}
+      path: ${{ steps.gto.outputs.path }}  # the path the model was `dvc get` to
```
aguschin, line 27:
```suggestion
      path: ${{ steps.gto.outputs.path }}
```

example-gto/code/mlem/.github/workflows/deploy-model-with-mlem.yml
```diff
@@ -16,7 +16,9 @@ jobs:
     - uses: actions/checkout@v3
     - name: "GTO: figure out what was registered/promoted and show the Registry state"
       id: gto
-      uses: iterative/gto-action@v1
+      uses: iterative/gto-action@v2
+      with:
+        download: true
```
aguschin, line 21:
```suggestion
```

amritghimire (approved):

PR #195, opened 2023-04-14
example-get-started-experiments: upgrade to dvclive>=2.5.0

dberenbaum:
Needed to support `log_artifact`

### DIFF
--- a/example-get-started-experiments/code/requirements.txt
+++ b/example-get-started-experiments/code/requirements.txt
@@ -1,5 +1,5 @@
 dvc[s3]>=2.52.0
-dvclive>=2.6.0
+dvclive>=2.5.0
 fastai
 python-box
 torch<2.0


example-get-started-experiments/code/requirements.txt
```diff
@@ -1,5 +1,5 @@
 dvc[s3]>=2.52.0
-dvclive>=2.6.0
```
daavoo, line 2: Wait but is 2.6.0 already pinned, not sure why the deployed project has a lower version

dberenbaum: Sorry, not sure how I missed that. Guess I need to pull.

daavoo: It looks like was my bad deploying from a branch that was not pinned or something because the live example doesn't contain this pin

PR #196, opened 2023-04-14
example-get-started-experiments: use cpu-friendly arch

dberenbaum:
Changed model arch to something that runs in reasonable time on cpu since I think it's not feasible for everyone trying it out to have access to gpu (or they may have some restricted access to one that they don't want to use to demo a new tool). Performance is not quite as strong but it's not a dramatic dropoff:

Screenshot 2023-04-14 at 9 52 12 AM

### DIFF
--- a/example-get-started-experiments/code/notebooks/TrainSegModel.ipynb
+++ b/example-get-started-experiments/code/notebooks/TrainSegModel.ipynb
@@ -45,7 +45,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


daavoo (commented):
@dberenbaum looks like the notebook needs some cleanup, the one pushed here contains errors, commented code, etc

Which is the architecture we should hardcode in the notebook?

daavoo:
BTW what is the runtime on your CPU?

dberenbaum:
> @dberenbaum looks like the notebook needs some cleanup, the one pushed here contains errors, commented code, etc

Sorry, fixed now. Thanks for checking.

> Which is the architecture we should hardcode in the notebook?

`shufflenet_v2_x2_0` (you should also see it in `params.yaml`).

> BTW what is the runtime on your CPU?

On my CPU, a single training takes <1 minute.

daavoo:
Will build and push

PR #197, opened 2023-04-17
example-get-started: Update src code

leonardcser:
The reason for the PR is to make the source code clearer for the `example-get-started` while keeping it as simple as possible. The actual code was left unchanged.

Here are my changes for a more understandable starter source code:
- Formatted files using black code style
- Update the Python source code to make it clearer
- Modularise code into functions
- Document functions

### DIFF
--- a/example-get-started/code/src/evaluate.py
+++ b/example-get-started/code/src/evaluate.py
@@ -11,21 +11,17 @@ from dvclive import Live
 from matplotlib import pyplot as plt
 
 
-EVAL_PATH = "eval"
+def evaluate(model, matrix, split, live, save_path):
+    """
+    Dump all evaluation metrics and plots for given datasets.
 
-
-if len(sys.argv) != 3:
-    sys.stderr.write("Arguments error. Usage:\n")
-    sys.stderr.write("\tpython evaluate.py model features\n")
-    sys.exit(1)
-
-model_file = sys.argv[1]
-train_file = os.path.join(sys.argv[2], "train.pkl")
-test_file = os.path.join(sys.argv[2], "test.pkl")
-
-
-def evaluate(model, matrix, split, live):
-    """Dump all evaluation metrics and plots for given datasets."""
+    Args:
+        model (sklearn.ensemble.RandomForestClassifier): Trained classifier.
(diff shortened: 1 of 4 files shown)


daavoo:
> what do you think @daavoo @dberenbaum ?

I think the `src` in this P.R. looks more like the state I would expect for people using a `dvc.yaml` with isolated scripts per stage.

> and it makes it farther away from a quite regular (for good or bad) notebook-style flow 

I think the motivation is not good enough for trying to make artificially "messy" scripts.

We could have an actual notebook in the project like in [example-get-started-experiments](https://github.com/iterative/example-repos-dev/tree/master/example-get-started-experiments) and acknowledge that you are expected to clean up and refactor the code when wanting to use `dvc.yaml` , like we acknowledge in https://dvc.org/doc/start/experiments/experiment-pipelines#stepping-up-and-out-of-the-notebook

daavoo:
> means that code is quite complicated for a DS project (should we try to simplify it?), 

Which parts look complicated? It doesn't look complicated for a DS project to me.

example-get-started/code/src/evaluate.py
```diff
@@ -11,21 +11,17 @@
...
-
-model_file = sys.argv[1]
-train_file = os.path.join(sys.argv[2], "train.pkl")
-test_file = os.path.join(sys.argv[2], "test.pkl")
-
-
-def evaluate(model, matrix, split, live):
-    """Dump all evaluation metrics and plots for given datasets."""
+def evaluate(model, matrix, split, live, save_path):
+    """
```
daavoo, line 15: These docstrings feel like not adding much context compared to potential type hints.

Not saying that we should drop it from the P.R. or extend them, just commenting that I would probably not add them in a real project.

leonardcser: Thanks for your feedback! While I understand your concern about the potential clutter that more documentation could bring, I think that providing clear and detailed docstrings can be incredibly helpful, especially for those who are new to Python or may not be familiar with this specific project libraries. Overall, I believe that more documentation is rarely a bad thing.

daavoo (approved):

PR #201, opened 2023-05-08
exp: drop aug_transforms for mac m1 compatibility

dberenbaum:
Drops the `aug_transforms` calls, which AFAIU were only being used to set the output size of the images. This was making the repo incompatible with mac m1, which works out of the box. Model still seems to perform fine without them.

### DIFF
--- a/example-get-started-experiments/code/src/train.py
+++ b/example-get-started-experiments/code/src/train.py
@@ -11,7 +11,6 @@ from fastai.metrics import DiceMulti
 from fastai.vision.all import (
     Resize,
     SegmentationDataLoaders,
-    aug_transforms,
     imagenet_stats,
     models,
     unet_learner,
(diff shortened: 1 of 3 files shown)


example-get-started-experiments/code/src/train.py
```diff
@@ -42,7 +41,6 @@ def train():
         valid_pct=params.train.valid_pct,
         item_tfms=Resize(params.train.img_size),
         batch_tfms=[
-            *aug_transforms(size=params.train.img_size),
```
shcheklein, line 45: not clear why it was needed at all tbh, if we the Resize above cc @alex000kim

alex000kim: Not really needed in this case and can be removed. In general, `aug_transforms` can do some transformations to the batch of images that can change their size (e.g. `RandomCrop`) from what was used earlier in `item_tfms`

alex000kim (approved):

example-get-started-experiments/code/src/train.py
```diff
@@ -42,7 +41,6 @@ def train():
         valid_pct=params.train.valid_pct,
         item_tfms=Resize(params.train.img_size),
```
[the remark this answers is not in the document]
dberenbaum: Good catch, added back the param.

PR #207, opened 2023-06-20
Add model and register versions

### LINKED ISSUE #203: Register models in the example repos to see them in the MR
Feedback by @omesser in one of the Studio tickets.

aguschin:
close #203

### DIFF
--- a/example-get-started-experiments/README.md
+++ b/example-get-started-experiments/README.md
@@ -52,7 +52,14 @@ and create a P.R. from the best of the 3 experiments that are found in the lates
 
 - Add a model to Studio Model Registry
 
-Go to Studio MR and click "Add a model". Fill the `path=models/model.pkl` and add a model to a separate branch. 
+Go to Studio MR and click "Add a model". Details:
+```
+name: pool-segmentation
+path: models/model.pkl
+description: "This is a Computer Vision (CV) model that solves the problem of segmenting out swimming pools from satellite images"
+labels: cv, satellite-images, segmentation
+```
+Fill the `path=models/model.pkl` and add a model to a separate branch. 
 Copy other details from the existing repo or fill them from scratch. 
 Register new version and assign `dev` stage to it.
 Open the public MR from studio.iterative.ai, find the model, copy URL to model details page.
--- a/example-get-started-experiments/README.md
+++ b/example-get-started-experiments/README.md
@@ -45,6 +45,10 @@ You may remove the generated repo with:
 rm -fR build
 ```
 
+Note that you may need to reparse the repo. Pay attention to whether experiments shown
+in experiments table are "detached" or if the experiments you just pushed doesn't
+show up in the Project table.
+
 - Manual Studio P.R.
 
 Once the repo has been generated and pushed, go to the 
--- a/example-get-started-experiments/README.md
+++ b/example-get-started-experiments/README.md
@@ -49,18 +49,3 @@ rm -fR build
 Once the repo has been generated and pushed, go to the 
 [corresponding Studio project](https://studio.iterative.ai/team/Iterative/projects/example-get-started-experiments-y8toqd433r) 
 and create a P.R. from the best of the 3 experiments that are found in the latest commit of `main` branch.
-
-- Add a model to Studio Model Registry
-
-Go to Studio MR and click "Add a model". Details:
-```
-name: pool-segmentation
-path: models/model.pkl
-description: "This is a Computer Vision (CV) model that solves the problem of segmenting out swimming pools from satellite images"
-labels: cv, satellite-images, segmentation
-```
-Fill the `path=models/model.pkl` and add a model to a separate branch. 
-Copy other details from the existing repo or fill them from scratch. 
-Register new version and assign `dev` stage to it.
-Open the public MR from studio.iterative.ai, find the model, copy URL to model details page.
-Post that url in the PR created by Studio.
--- a/example-get-started-experiments/README.md
+++ b/example-get-started-experiments/README.md
@@ -15,7 +15,8 @@ Run these commands to force push it:
 cd build/example-get-started-experiments
 git remote add origin https://github.com/iterative/example-get-started-experiments.git
 git push --force origin main
-git push --force origin --tags
+# we push git tags one by one for Studio to receive webhooks:
+git tag --sort=creatordate | xargs -n 1 git push --force origin
 ```
 
 Run these to drop and then rewrite the experiment references on the repo:
--- a/example-get-started-experiments/code/src/train.py
+++ b/example-get-started-experiments/code/src/train.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import numpy as np
 import torch
 from box import ConfigBox
+from dvclive import Live
 from dvclive.fastai import DVCLiveCallback
 from fastai.data.all import Normalize, get_files
 from fastai.metrics import DiceMulti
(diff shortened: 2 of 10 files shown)


example-get-started/generate.sh
```diff
@@ -135,6 +143,20 @@ dvc stage add -n train \
   -o model.pkl \
   python src/train.py data/features model.pkl
 dvc repro
+python <<EOF
+from dvc.repo import Repo
```
shcheklein, line 147: feel we need a command for that? :)

may be it's easier append to dvc.yaml via bash?

aguschin: yes, we need that command I guess!

Let's use python approach if no strong objections.

shcheklein: yep, no strong objections at all. Can use python, no worries. Thanks @aguschin for updating this repo.

example-get-started/code/src/featurization.py
```diff
@@ -56,7 +56,7 @@ def generate_and_save_train_features(train_input, train_output, bag_of_words, tf
         tfidf (sklearn.feature_extraction.text.TfidfTransformer): TF-IDF transformer.
     """
     df_train = get_df(train_input)
-    train_words = np.array(df_train.text.str.lower().values.astype("U"))
+    train_words = np.array(df_train.text.str.lower().values)
```
shcheklein, line 59: curious - why this change?

shcheklein (approved):
just one comment, otherwise should be good to go?

example-get-started/generate.sh
```diff
@@ -135,6 +143,20 @@ dvc stage add -n train \
...
+repo = Repo(".")
+artifact = Artifact(
+  path="model.pkl", 
+  type="model",
+  desc="Detect whether the given stackoverflow question should have R language tag",
+  labels=["nlp", "classification", "stackoverflow"]
+)
+repo.artifacts.add("text-classification", artifact)
+EOF
+
```
dberenbaum, line 159: Why not change the source code to use `live.log_artifact()`?

example-get-started/code/src/featurization.py
aguschin: It was breaking (just terminating the process without any hints). Not sure what was the problem. And IIUC, `.astype("U")` convert value to string, but it's already string. So not sure why it was needed in the first place :)

example-get-started/generate.sh
aguschin: Right 😂 Not sure how I missed that. Will do!

example-get-started/code/src/featurization.py
shcheklein: @aguschin can we check blame please and see who put it there in the first place and ask for the review? (can be me of course :) just in case if @daavoo and @dberenbaum remember something)

aguschin: This goes all the way back to the creation of this repo 5 years ago: https://github.com/iterative/example-repos-dev/blame/e22378ce6a7c1729d3d5bdfe365be38bad605353/code/featurization.py

it could be python 2 (for which this operation made sense IIUC)

don't think it's relevant to double-check this :)

aguschin:
For the record, dvclive's `log_artifact` doesn't add artifact to `dvc.yaml` right now. @daavoo is looking into it.

example-get-started-experiments/README.md
```diff
@@ -52,7 +52,14 @@ and create a P.R. from the best of the 3 experiments that are found in the lates
 
 - Add a model to Studio Model Registry
 
-Go to Studio MR and click "Add a model". Fill the `path=models/model.pkl` and add a model to a separate branch. 
+Go to Studio MR and click "Add a model". Details:
```
shcheklein, line 55: it's not related?, let's do a separate PR please

example-get-started/generate.sh
```diff
@@ -63,7 +63,16 @@ git tag -a "1-dvc-init" -m "DVC initialized."
 mkdir data
 dvc get https://github.com/iterative/dataset-registry \
   get-started/data.xml -o data/data.xml
-dvc add data/data.xml --desc "Initial XML StackOverflow dataset (raw data)"
+python <<EOF
+from dvc.repo import Repo
```
shcheklein, line 67: can we also use `log_artifact`?

example-get-started/generate.sh
```diff
@@ -170,9 +178,11 @@ echo "plots:
 dvc repro
 git add .gitignore dvc.yaml dvc.lock eval
 tick
-git commit -m "Create evaluation stage"
+git commit -am "Create evaluation stage"
 git tag -a "8-evaluation" -m "Baseline evaluation stage created."
 git tag -a "baseline-experiment" -m "Baseline experiment evaluation"
+gto register dvclive:text-classification --version v1.0.0
```
shcheklein, line 184: do we use dvclive with a custom path in the eval stage? let's try to reuse it ... I don't think we want to have the `dvclive` dir in this project ... or at least is should be a single dir with everything in it

daavoo:
I am ok with going with the P.R. (after addressing the comments) but it feels to me that we need to take a step back and discuss the "Model Registry using DVCLive log_artifact" workflow.

I didn't play with it until now because there was no Studio support, but I have now tried to set up an example myself (and reviewed this P.R.) and the experience feels unpolished.

example-get-started/generate.sh
aguschin: We can, but I don't think it's logical (it's either a separate python script or modifying `prepare.py`, which doesn't make sense to me). Rather I can simply echo few lines from bash if that's better.

example-get-started/generate.sh
aguschin: Looks somewhat strange to me to have a model registered in `eval/live` TBH. For this particular case, it be good to write to `dvc.yaml` in the root, e.g. using
```
with Live("") as live:
  ...
```
Wdyt @daavoo?

example-get-started/generate.sh
aguschin: wdyt?

example-get-started/generate.sh
daavoo: > it be good to write to dvc.yaml in the root, e.g. using

I think this would not work or would work in unexpected ways.

We could update the repo to have a parent  `results` output folder and use `results/train` and `results/eval` in each stage, I think it is what Ivan was suggesting

aguschin: Looks like a potentially substantial update to me (I assume we should move `model.pkl` to `results/train` as well?). I suggest to keep this repo simple. Let's use the approach with `results/...` for the `example-gs-experiments` repo and here simply edit `dvc.yaml` in the root like before by calling DVC API.

aguschin: so unless you're against, I'm reverting the changes.

aguschin:
This is how it looks like now:
image

aguschin:
ok, I re-generated both repos and pushed changes. In Studio everything looks good AFAIT.

@daavoo, can I ask you to review changes for example-get-started-experiments? Need a second look in case I missed something. Thanks!

example-get-started-experiments/code/src/train.py
```diff
@@ -55,17 +56,25 @@ def train():
...
-    learn.fine_tune(
-        **params.train.fine_tune_args,
-        cbs=[DVCLiveCallback(dir="results/train", report="md")],
-    )
-    models_dir = Path("models")
-    models_dir.mkdir(exist_ok=True)
-    learn.export(fname=(models_dir / "model.pkl").absolute())
+        learn.fine_tune(
+            **params.train.fine_tune_args,
+            cbs=[DVCLiveCallback(dir="results/train", report="md")],
```
daavoo, line 66:
```suggestion
            cbs=[DVCLiveCallback(live=live)],
```
https://dvc.org/doc/dvclive/ml-frameworks/fastai#examples

PR #208, opened 2023-06-21
Bring back commit-based CI

daavoo:
This is the current deployed version in https://github.com/iterative/example-get-started-experiments

image

### DIFF
--- a/example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
+++ b/example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
@@ -61,5 +61,12 @@ jobs:
           dvc exp run
           dvc remote add --local push_remote s3://dvc-public/remote/get-started-pools 
           dvc push -r push_remote
+
           cml pr --squash --skip-ci .
-          cml comment create --pr --publish results/evaluate/report.md
+          
+          echo "## Metrics" >> report.md
+          dvc metrics diff main --md >> report.md
+          echo "## Params" >> report.md
+          dvc params diff main --md >> report.md
+
+          cml comment create --pr report.md
--- a/example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
+++ b/example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
@@ -1,17 +1,19 @@
-name: DVC Studio Experiment
+name: Run Studio Experiment
 on: 
-  workflow_dispatch:
-    inputs:
-      exp-run-args:
-        description: 'Args to be passed to dvc exp run call'
-        required: false
-        default: ''
+  push:
+    paths:
+      - ".github/workflows/run-experiment.yaml"
+      - "data/**"
+      - "src/**"
+      - "params.yaml"
+      - "dvc.*"
 permissions:
   contents: write
   id-token: write
   pull-requests: write
 jobs:
   deploy-runner:
+    if: github.actor == 'iterative-studio[bot]'
     environment: cloud
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
```diff
@@ -1,17 +1,19 @@
-name: DVC Studio Experiment
+name: Run Studio Experiment
 on: 
-  workflow_dispatch:
-    inputs:
-      exp-run-args:
-        description: 'Args to be passed to dvc exp run call'
-        required: false
-        default: ''
+  push:
```
shcheklein, line 3: we want probably to exclude the main branch?

example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
```diff
@@ -1,17 +1,19 @@
...
+      - "src/**"
+      - "params.yaml"
+      - "dvc.*"
 permissions:
   contents: write
   id-token: write
   pull-requests: write
 jobs:
   deploy-runner:
+    if: github.actor == 'iterative-studio[bot]'
```
shcheklein, line 16: ok, this should do the trick with running it on Studio updates only

example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
```diff
@@ -54,7 +56,17 @@ jobs:
...
+          dvc push -r push_remote
+
+          cml pr --squash --skip-ci .
+          
+          echo "## Metrics" >> report.md
+          dvc metrics diff main --md >> report.md
+          echo "## Params" >> report.md
+          dvc params diff main --md >> report.md
+
+          cml comment create --pr report.md
```
shcheklein, line 72: eof below 👇

shcheklein, line 72: eof above ⬆️

example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
```diff
@@ -54,7 +56,17 @@ jobs:
...
-          dvc remote add --local push_remote s3://dvc-public/remote/get-started-pools
-          dvc exp run --pull --allow-missing ${{ github.event.inputs.exp-run-args }} -v
-          dvc exp push origin --rev HEAD -v
+          dvc exp run
+          dvc remote add --local push_remote s3://dvc-public/remote/get-started-pools 
+          dvc push -r push_remote
+
+          cml pr --squash --skip-ci .
+          
+          echo "## Metrics" >> report.md
```
shcheklein, line 67: minor: can be `>` instead of `>>` to be safe

example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
```diff
@@ -54,7 +56,17 @@ jobs:
           DVCLIVE_LOGLEVEL: DEBUG
```
shcheklein, line 56: do we need Studio repo url?

example-get-started-experiments/code/.github/workflows/runs-studio-experiment.yml
```diff
@@ -1,17 +1,19 @@
-name: DVC Studio Experiment
+name: Run Studio Experiment
 on: 
-  workflow_dispatch:
```
shcheklein, line 3: may be still keep it? (can demo that is also works nicely?)

shcheklein: and makes it easier to debug ...

shcheklein (approved):
Good, thanks @daavoo . Some minor comments.

daavoo:
Pushed a new version with the comments.
Also brought back the workflow dispatch and tested that it works (there are some bugs with detached experiments but it looks it is on Studio side, not in the workflow)

PR #209, opened 2023-06-22
More CI workflow updates

daavoo:
- Use `on push` trigger

Using `pull_request` requires modifying the action/checkout step in order to properly set the git HEAD (which was causing the experiments to appear as detached in Studio)

on push has the same effect and we use the github actor to filter events

- Use `exp push` for sharing in workflow dispatch mode.

You can manually replicate from github UI the future Studio UI behavior

### DIFF
--- a/example-get-started-experiments/code/.github/workflows/run-experiment.yml
+++ b/example-get-started-experiments/code/.github/workflows/run-experiment.yml
@@ -59,15 +59,14 @@ jobs:
 
       - run: pip install -r requirements.txt
 
-      - name: Trrain
+      - name: Train
         env:
           REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
           STUDIO_TOKEN: ${{ secrets.STUDIO_TOKEN }}
           DVCLIVE_LOGLEVEL: DEBUG
         run: |
           cml ci --fetch-depth 0
-          dvc pull
-          dvc exp run ${{ github.event.inputs.exp-run-args }}
+          dvc exp run --pull --allow-missing ${{ github.event.inputs.exp-run-args }}
           dvc remote add --local push_remote s3://dvc-public/remote/get-started-pools 
 
       - name: Workflow Dispatch Sharing
@@ -75,7 +74,7 @@ jobs:
         env:
           STUDIO_TOKEN: ${{ secrets.STUDIO_TOKEN }}
         run: |
-          dvc exp push origin --rev HEAD -r push_remote
+          dvc exp push origin -r push_remote
 
       - name: Commit-based Sharing
         if: github.actor == 'iterative-studio[bot]'
--- a/example-get-started-experiments/code/.github/workflows/run-experiment.yml
+++ b/example-get-started-experiments/code/.github/workflows/run-experiment.yml
@@ -1,15 +1,14 @@
 name: Run Experiment
 on: 
+
+  push:
+
   workflow_dispatch:
     inputs:
       exp-run-args:
         description: 'Args to be passed to dvc exp run call'
         required: false
         default: ''
-
-  pull_request:
-    branches:    
-      - main
 
 permissions:
   contents: write
@@ -22,6 +21,7 @@ jobs:
     if: ${{ (github.actor == 'iterative-studio[bot]') || (github.event_name == 'workflow_dispatch') }} 
     environment: cloud
     runs-on: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v3
       - uses: iterative/setup-cml@v1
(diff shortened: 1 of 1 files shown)


daavoo:
Ok there is one more thing I want to do for workflow dispatch

example-get-started-experiments/code/.github/workflows/run-experiment.yml
```diff
@@ -64,32 +65,27 @@ jobs:
           STUDIO_TOKEN: ${{ secrets.STUDIO_TOKEN }}
```
dberenbaum, line 65: There's a type above in the name: `Trrain`->`Train`

example-get-started-experiments/code/.github/workflows/run-experiment.yml
```diff
@@ -64,32 +65,27 @@ jobs:
...
           DVCLIVE_LOGLEVEL: DEBUG
         run: |
-          dvc pull
           cml ci --fetch-depth 0
```
dberenbaum, line 68: No action needed, but we only need `--fetch-depth 0` for the report step below and will eventually drop it, right? I think another option would be to add `git fetch` at the start of the report step.

example-get-started-experiments/code/.github/workflows/run-experiment.yml
```diff
@@ -64,32 +65,27 @@ jobs:
...
+          dvc pull
```
dberenbaum, line 69: Any reason for this instead of `dvc exp run --pull --allow-missing`?

example-get-started-experiments/code/.github/workflows/run-experiment.yml
```diff
@@ -64,32 +65,27 @@ jobs:
...
           dvc exp run ${{ github.event.inputs.exp-run-args }}
           dvc remote add --local push_remote s3://dvc-public/remote/get-started-pools
```
dberenbaum, line 71: Not really actionable, but we should try to get back to enabling [anonymous login](https://github.com/iterative/dvc/issues/5797) and then add this as secondary project remote since now it seems a bit awkward.

example-get-started-experiments/code/.github/workflows/run-experiment.yml
```diff
@@ -64,32 +65,27 @@ jobs:
...
 
-      - name: Workflow Dispatch PR
+      - name: Workflow Dispatch Sharing
         if: github.event_name == 'workflow_dispatch'
         env:
-          REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+          STUDIO_TOKEN: ${{ secrets.STUDIO_TOKEN }}
         run: |
-          cml pr --skip-ci .
+          dvc exp push origin --rev HEAD -r push_remote
```
dberenbaum, line 78:
```suggestion
          dvc exp push origin -r push_remote
```

Can we drop it now? By the way, not sure why we weren't recommending the simpler `-A` here anyway.

example-get-started-experiments/code/.github/workflows/run-experiment.yml
daavoo: The fix was not released at the point I used this. Will update

example-get-started-experiments/code/.github/workflows/run-experiment.yml
daavoo: Yep! was also not released when I pushed this

example-get-started-experiments/code/.github/workflows/run-experiment.yml
dberenbaum: And I mistyped `typo` 🤦

example-get-started-experiments/code/.github/workflows/run-experiment.yml
daavoo: > and will eventually drop it, right?

Yes, I hope to drop it soon in favor of a Studio-based one.

dberenbaum (approved):

example-get-started-experiments/code/.github/workflows/run-experiment.yml
daavoo: Also, would there be any harm in having the push_remote already set up in the `.dvc/config`?

dberenbaum: Right, that's what I was thinking, but that's also awkward without anonymous login since it can't be used by anyone external. Anyway, I'm fine with it either way for now.

shcheklein: I'm fine (and do this in the hacktathon repo) to have the second remote in the config. I think it's even better (can educate ppl about different remotes,. etc)

PR #212, opened 2023-06-30
Updates to example-get-started-experiments

### LINKED ISSUE #188: `experiments`: Use fixed test set
We are doing a random split so whenever we change the `data_split` stage (i.e adding more data or changing the percentage), we evaluate on a different set of images and generate different results. 

This is a problem because the comparison of metrics is invalid as they come from different datasets.
It has also a minor but a more visual problem that we can't compare predictions in Studio:

Captura de pantalla 2023-02-16 a las 13 28 30

daavoo:
- Push to private repos used by Studio testing.
- Deactivate/Ignore DVCLive reports.
- Use cache_images in DVCLive.
- Use a fixed test set. Closes #188


daavoo:
I am getting errors when pushing to AWS from codespaces, not sure if something changed @shcheklein

shcheklein:
no, nothing has changed. Let me try ...

shcheklein:
@daavoo can you run `aws s3 ls` - does it work for you? what is the path you are trying to push to and what is the error you are getting?

daavoo:
> @daavoo can you run `aws s3 ls` - does it work for you? what is the path you are trying to push to and what is the error you are getting?

Ok, I know what happened now that I re-logged into the codespace again. It looks like I missed a prompt when you log in that asks about setting the default AWS region.
After setting that, it works

example-get-started-experiments/generate.sh
```diff
@@ -117,8 +117,6 @@ dvc stage add -n evaluate \
   -d src/evaluate.py -d models/model.pkl -d data/test_data \
   python src/evaluate.py
 
-rm results/train/dvc.yaml
-git rm --cached results/train/dvc.yaml
```
dberenbaum, line 121: What's this change about?

daavoo: It was unnecessary. I don't recall why it was there. I assume that in an older version, we used to define the metrics and plots in the top level dvc.yaml so this was introduced to prevent duplication

dberenbaum (approved):

example-get-started-experiments/code/.gitignore
```diff
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# DVCLive report
+dvclive/report.md
```
shcheklein, line 132: newline

.devcontainer.json
```diff
@@ -20,6 +20,12 @@
         },
         "iterative/example-get-started-experiments": {
           "permissions": "write-all"
+        },
+        "iterative/example-get-started-private": {
+          "permissions": "write-all"
+        },
+        "iterative/example-get-started-experiments-private": {
```
shcheklein, line 27: how do we use the private repo?

daavoo: For testing in Studio. Updated name

Issue #200, opened 2023-04-19
Automate deploy


daavoo:
We currently rely on people generating the repo locally or in codespaces and then manually running the commands described (i.e. https://github.com/iterative/example-repos-dev/blob/master/example-get-started-experiments/README.md) in order to "deploy" the new version of the repo.

At least for [`example-get-started-experiments`](https://github.com/iterative/example-repos-dev/tree/master/example-get-started-experiments), which is the one I am familiar with, I don't see any reason why we should not automate this process in a GitHub workflow on `push: main` that generates and deploys the model (we can set up a CML runner for the GPU)

daavoo:
> , I don't see any reason why we should not automate this process 

Any downsides/thoughts @shcheklein @dberenbaum ?

shcheklein:
No downsides. There are some steps that needs to be taken care of on the Studio side sometimes (reparse, adjust Views columns that we need to keep an eye on). Also, primarily the reason was that we don't have time for this, it was not considered critical.

daavoo:
btw this would be also good to ensure that the deployed version is up to date.
It looks like the current https://github.com/iterative/example-get-started-experiments was deployed from a branch that was not up to date with master

dberenbaum:
Prioritizing this for a couple reasons:

1. Sick of spending time manually deploying or fixing issues deploying
2. Means we can deploy more frequently and ensure it works with newer versions of dvc

PR #213, opened 2023-07-03
Copy gitlab file

daavoo:


daavoo:
This is now deployed

dberenbaum:
Sorry, this is another reason we need #200.

It looks like the pipeline is failing in https://gitlab.com/iterative.ai/example-get-started-experiments/-/pipelines. Let's figure that out before we merge this in case there are other changes needed.

daavoo:
Well actually now looks like one triggered by you also failed 😫

dberenbaum (approved):

dberenbaum:
It's an issue with https://github.com/iterative/itops/pull/1799. It's working now on `master`, so the permissions just need to be updated to work on other branches.

PR #214, opened 2023-07-04
feat(example-get-started): config to generate nested repos

shcheklein:
Addresses partially https://github.com/iterative/studio/issues/6559 .

Add a way to generate repos similar to these ones:

Screenshot 2023-07-03 at 8 52 16 PM

### DIFF
--- a/example-get-started-experiments/README.md
+++ b/example-get-started-experiments/README.md
@@ -1,6 +1,6 @@
 Generate the actual repo by running: 
 
-```
+```shell
 bash generate.sh
 ```
 
@@ -11,36 +11,25 @@ permissions.
 
 Run these commands to force push it:
 
-```
+```shell
 cd build/example-get-started-experiments
-git remote add origin https://github.com/iterative/example-get-started-experiments.git
-git remote add fixture https://github.com/iterative/example-get-started-experiments-frozen.git
+git remote add origin git remote add origin git@github.com:<slug>/example-get-started-experiments.git
 git push --force origin main
-git push --force fixture main
 # we push git tags one by one for Studio to receive webhooks:
 git tag --sort=creatordate | xargs -n 1 git push --force origin
-git tag --sort=creatordate | xargs -n 1 git push --force frozen
 ```
 
 Run these to drop and then rewrite the experiment references on the repo:
 
-```
+```shell
 source .venv/bin/activate
 dvc exp remove -A -g origin
-dvc exp remove -A -g fixture
 dvc exp push origin -A
-dvc exp push fixture -A
-```
-
-And this to clean the remote cache to only contain the last iteration:
-
-```
-dvc gc -c --all-commits --all-experiments
 ```
 
 To push a copy to GitLab:
 
-```
+```shell
 git remote add gitlab git@gitlab.com:iterative.ai/example-get-started-experiments.git
 git push --force gitlab main
 # we push git tags one by one for Studio to receive webhooks:
@@ -49,24 +38,30 @@ git tag --sort=creatordate | xargs -n 1 git push --force gitlab
 dvc exp remove -A -g gitlab
 dvc exp push gitlab -A
 ```
+
 Finally, return to the directory where you started:
 
-```
+```shell
 cd ../..
 ```
 
 You may remove the generated repo with:
 
-```
+```shell
 rm -fR build
 ```
 
-Note that you may need to reparse the repo. Pay attention to whether experiments shown
-in experiments table are "detached" or if the experiments you just pushed doesn't
-show up in the Project table.
+To update the project in Studio, follow the instructions at:
 
-- Manual Studio P.R.
+https://github.com/iterative/studio/wiki/Updating-and-synchronizing-demo-project
+
+
+Pay attention to whether experiments shown in experiments table are "detached"
+or if the experiments you just pushed doesn't show up in the Project table.
+
+Manual Studio PR:
 
 Once the repo has been generated and pushed, go to the 
 [corresponding Studio project](https://studio.iterative.ai/team/Iterative/projects/example-get-started-experiments-y8toqd433r) 
-and create a P.R. from the best of the 3 experiments that are found in the latest commit of `main` branch.
+and create a PR from the best of the 3 experiments that are found in the latest
+commit of the `main` branch.
--- a/example-get-started/README.md
+++ b/example-get-started/README.md
@@ -1,0 +1,140 @@
+A set of scripts to generate an NLP DVC Studio project with multiple branches,
+commit history, experiments, metrics, plots, etc. It used in the DVC docs and in
+Studio as a demo project. 
+
+This script can be also used in an advanced scenario to generate a nested
+mono-repositories that are used as fixtures in Studio testing, or testing
+different remote types. See the section below for the advanced settings.
+
+## Demo project
+
+Note! In some cases, before rebuilding the project you might want to delete the
+existing remote tags if you change the order, or names.
+
+```shell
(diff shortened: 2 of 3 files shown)


daavoo (commented):
Can we add some instructions in the README? 

- How are these options supposed to be used?
- Are we expected to force-push those repos in the screenshot whenever the "main" example-get-started is regenerated?

shcheklein:
> Can we add some instructions in the README?

Yep. I plan to add the instruction in the repo itself that is being generated (and already did some steps in that direction - it dumps its own config)  + comments in the script. WDYT?

> Are we expected to force-push those repos in the screenshot whenever the "main" example-get-started is regenerated?

Yes, it's will be a force push or we can recreate them. We'll see. Not sure yet. Need to get more data on the usage first tbh.

shcheklein:
I've updated it one more time. One item left - to move the instructions into README. 

At the end it generates this kind of repos https://github.com/iterative/monorepo-fixture 

@daavoo @amritghimire @aguschin I suggest the we rename all the repos into <something>-fixture . I don't like `test` since it is sometimes someone actually testing it. It should be nonetheless visible that this is a fixed testing repo. 

Let's rename all those fixtures please one by one and use the same convention. Also as I mentioned somewhere, make sure that we can generate any of them anytime.

example-get-started/generate.sh
```diff
@@ -12,11 +12,64 @@ REPO_NAME="example-get-started"
...
+  fi
+
+  if [ $REMOTE == 'private-s3' ]; then
+    # Remote active on this env only, for writing to HTTP redirect below.
+    dvc remote add -d storage s3://dvc-private/remote/get-started
+  fi
+
+  if [ $REMOTE == 'private-http' ]; then
+    dvc remote add -d --local storage ssh://dvc@35.194.53.251/home/dvc/storage
+    dvc remote modify --local storage keyfile /Users/ivan/.ssh/dvc_gcp_remotes_rsa
```
daavoo, line 62: This makes it only generable by you.
I guess it is ok if it is not mean to be generated frequently

daavoo (approved):
Approving to not block the merge. I will update `-fixture` name in the experiments one

example-get-started/generate.sh
shcheklein: > This makes it only generable by you.

Yep, I'll add the details to the README for this repo. Ppl can manually change the path for now, not critical.

example-get-started-experiments/README.md
```diff
@@ -24,20 +24,14 @@ git tag --sort=creatordate | xargs -n 1 git push --force frozen
...
-```
+```shell
 source .venv/bin/activate
 dvc exp remove -A -g origin
 dvc exp remove -A -g fixture
 dvc exp push origin -A
 dvc exp push fixture -A
 ```
 
-And this to clean the remote cache to only contain the last iteration:
```
daavoo, line 35: Whi dropping this?

example-get-started/README.md
```diff
@@ -0,0 +1,127 @@
...
+Run these commands to force push it:
+
+```shell
+cd build/example-get-started
+git remote add origin git@github.com:<slug>/example-get-started.git
+git push --force origin main
+git push --force origin try-large-dataset
+git push --force origin tune-hyperparams
+# we push git tags one by one for Studio to receive webhooks:
+git tag --sort=creatordate | xargs -n 1 git push --force origin
```
daavoo, line 53:
```suggestion
cd build/example-get-started
git remote add origin git@github.com:iterative/example-get-started.git
git remote add fixture git@github.com:iterative/example-get-started-fixture.git
git push --force origin main
git push --force fixture main
git push --force origin try-large-dataset
git push --force fixture try-large-dataset
git push --force origin tune-hyperparams
git push --force fixture tune-hyperparams
# we push git tags one by one for Studio to receive webhooks:
git tag --sort=creatordate | xargs -n 1 git push --force origin
git tag --sort=creatordate | xargs -n 1 git push --force fixture
```

amritghimire (approved):

example-get-started/README.md
aguschin: I guess we need some note here that after you push this, tests in Studio can be broken. Should probably be expected by everyone, but better to have a reminder you need to check them and fix if needed.

example-get-started/README.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+Run these to drop and then rewrite the experiment references on the repo:
+
+```shell
+source .venv/bin/activate
+dvc exp remove -A -g origin
+dvc exp push origin -A
```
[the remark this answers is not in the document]
aguschin: Looks like there's an option `OPT_TESTING_REPO` that will generate a different repo (README is different). Is this expected, that in that case you'll be pushing all refs to origin or fixture repos?

example-get-started/README.md
aguschin: Also, from my user perspective, I'd want to first push everything to example-get-started, then go check everything is correct (open View, MR, MDP), and then want to push to fixture repo. So IMHO it's more convenient to have this separated.

daavoo: > Also, from my user perspective, I'd want to first push everything to example-get-started, then go check everything is correct (open View, MR, MDP), and then want to push to fixture repo. So IMHO it's more convenient to have this separated.

To my mind, the "right" process would be that you test the things in a personal copy or something, then when you have verified that everything is ok, merge the P.R to `main` in this repo, and everything is automatically generated/pushed to the right repos

daavoo: We are missing the automation in `main` branch

aguschin: Right, but how you do that with codespaces? Are you able to push to a repo in a personal account?

aguschin: Though it probably can be resolved with running `gh auth`.

example-get-started-experiments/README.md
shcheklein: it's dangerous - we use the same remote across multiple repos (e.g. fixture and non-fixture, etc).

example-get-started/README.md
shcheklein: folks, I think we are overcomplicating this ... we can mention that we also have fixtures (and in this case we'll have 5-6 of them - we can't in one README include all the details on how to generate them). Each fixture repo should include that in its readme (and that's what I do here)+ we should have a list of them somewhere (I'll start on that after I get through monorepo and MR tests - I hit bugs even writing tests, that's why it takes longer)

example-get-started/README.md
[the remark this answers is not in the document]
shcheklein: yes, it's intentional - in a fixture repo README contains the details on how generate it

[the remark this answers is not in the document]
shcheklein: so, no, these single instruction - generate once - push everywhere doesn't work in most cases

example-get-started/generate.sh
```diff
@@ -268,55 +405,9 @@ unset GIT_AUTHOR_EMAIL
...
-You may remove the generated repo with:
-
-rm -fR build
-
-`"
+set +eux
+echo
+echo "=========================================="
+echo "Done! Read README for the next steps."
+echo "=========================================="
```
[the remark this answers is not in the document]
shcheklein: Yep, I think it became non practical (Markdown markup, length, all the details about monorepo, etc) ... easier to read it

example-get-started/generate.sh
shcheklein: added the README and link to the Slack thread with the credentials

shcheklein:
Merging this, but we can keep the discussion and improving docs / READMEs on the fixtures gen.

PR #220, opened 2023-07-11
Rename run-experiment to dvc-studio. Add cml runner inputs.

daavoo:
Per https://github.com/iterative/studio/pull/6755

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -21,10 +21,10 @@
         "iterative/example-get-started-experiments": {
           "permissions": "write-all"
         },
-        "iterative/example-get-started-frozen": {
+        "iterative/example-get-started-fixture": {
           "permissions": "write-all"
         },
-        "iterative/example-get-started-experiments-frozen": {
+        "iterative/example-get-started-experiments-fixture": {
           "permissions": "write-all"
         }
       }
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -20,12 +20,6 @@
         },
         "iterative/example-get-started-experiments": {
           "permissions": "write-all"
-        },
-        "iterative/example-get-started-fixture": {
-          "permissions": "write-all"
-        },
-        "iterative/example-get-started-experiments-fixture": {
-          "permissions": "write-all"
         }
       }
     }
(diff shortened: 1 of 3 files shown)


daavoo:
I need to actually generate and push the repos

shcheklein (approved):

example-get-started-experiments/README.md
```diff
@@ -13,17 +13,21 @@ Run these commands to force push it:
 
 ```shell
 cd build/example-get-started-experiments
-git remote add origin git remote add origin git@github.com:<slug>/example-get-started-experiments.git
+git remote add origin https://github.com/iterative/example-get-started-experiments.git
+git remote add fixture https://github.com/iterative/example-get-started-experiments-fixture.git
```
shcheklein, line 17: @daavoo I would vote against this (using iterative directly here and sharing a command for the fixture update). We'll have a bunch of fixtures, we need also to start versioning them (like https://github.com/iterative/monorepo-fixture-v1 so that we don't break the existing tests a we migrate). This way it's easier to break things by copy pasting everything and running it w/o doing a check (e.g. replace iterative -> shcheklein to test things).

.devcontainer.json
```diff
@@ -21,10 +21,10 @@
         "iterative/example-get-started-experiments": {
           "permissions": "write-all"
         },
-        "iterative/example-get-started-frozen": {
+        "iterative/example-get-started-fixture": {
           "permissions": "write-all"
         },
-        "iterative/example-get-started-experiments-frozen": {
+        "iterative/example-get-started-experiments-fixture": {
```
shcheklein, line 27: do we need it to be able to generate live experiments? I'm a bit concerned since I think we need to start using versioning for the fixture (to avoid breaking tests)

daavoo: I think it is actually ok to drop it. As long as I have GPU to generate the "demo" one, I can do the push to this (or versioned) fixture by fetching the new demo.

example-get-started-experiments/README.md
daavoo: Makes sense. I will update the instructions.
I think it would still be nice to have this run automatically on CI for the "demo" one

PR #221, opened 2023-07-12
feat(example-get-started): add options needed for model registry mono…

shcheklein:
Continuation of https://github.com/iterative/studio/issues/6559

To unskip the last remaining skipped real parsing test this is needed to generate a fixture like https://github.com/iterative/monorepo-model-registry-fixture

### DIFF
--- a/example-get-started/generate.sh
+++ b/example-get-started/generate.sh
@@ -17,9 +17,13 @@ OPT_INIT_GIT='true' # Default true.
 OPT_INIT_DVC='true' # Default true.
 OPT_NON_DVC='false' # Default false.
 OPT_BRANCHES='true' # Default true.
-OPT_REMOTE="public-s3" # Default "public-s3". Other options: "public-s3", "private-http", "private-ssh", etc.
+OPT_TAGS='false' # Default true.
+OPT_REMOTE='public-s3' # Default "public-s3". Other options: "public-s3", "private-http", "private-ssh", etc.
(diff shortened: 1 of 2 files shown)


example-get-started/generate.sh
```diff
@@ -331,8 +342,8 @@ EOF
   dvc repro train
   tick
   git commit -am "${COMMIT_PREFIX}Reproduce model using bigrams"
-  git tag -a "9-bigrams-model${GIT_TAG_SUFFIX}" -m "Model retrained using bigrams."
-  if [ $OPT_REGISTER_MODELS == "true" ]; then
+  create_tag "9-bigrams-model${GIT_TAG_SUFFIX}" "Model retrained using bigrams."
+  if [ $OPT_TAG_MODELS == "true" ]; then
     gto register "${GTO_PREFIX}text-classification" --version v1.1.0
```
daavoo, line 347: Shouldn't this be also updated to use `OPT_MODEL_NAME` instead of `text-classification`?

daavoo (approved):
Minor comment

example-get-started/generate.sh
shcheklein: that's a really good catch! thanks a lot @daavoo 🙏  I'll update it and will update the tags in the new repo I generated with it (I'm hitting some issues on BE that @amritghimire has handled already, after that I'll create PR with a migration to this new fixture)

PR #226, opened 2023-07-22
fix(example-get-started): avoid inf and nans in prc plot output

### LINKED ISSUE #225: `example-get-started`: DVC plots fail to parse output with `SyntaxError: Unexpected token I in JSON`
Context: https://github.com/iterative/example-get-started/actions/runs/5484365547/job/14852022942?pr=53

Specifically:

```
SyntaxError: Unexpected token I in JSON at position 121
    at JSON.parse (<anonymous>)
    at /usr/local/lib/node_modules/vega-lite/bin/render.js:66:31
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Error: Process completed with exit code 1.
```

When it executes:

```
 dvc plots diff $PREVIOUS_REF workspace \
    --show-vega --targets Precision-Recall > vega.json
  vl2svg vega.json prc.svg
```

It "easy" to reproduce it locally with `vl2svg` installed via `npm install canvas@2 vega@5 vega-cli@5 vega-lite@5` (on Mac OS you would need to use `nvm` to force node x64 vs arm64, e.g. https://stackoverflow.com/a/70572866/298182).

The reason is that since https://github.com/scikit-learn/scikit-learn/pull/26194/files PRC produces `Infinity` in JSON. That Vega CLI tool fails to parse. Online / browser and I think VS Code work fine with that.

shcheklein:
Fixes #225 

Since we have PRs broken in the example repo https://github.com/iterative/example-get-started/actions/runs/5484365547/job/14852022942?pr=53 and I failed to demo it the last week.

It's a workaround for now. The best fix it work with Vega to fix their CLI. I'm not sure it's a priority at the moment since we have quick workaround if needed, JS version of Vega seems to work fine, and I'm not sure for the long term future of Vega in the project (tbh we need to migrate I think to something else after all as soon as we have more time and resources).

### DIFF
--- a/example-get-started/code/src/evaluate.py
+++ b/example-get-started/code/src/evaluate.py
@@ -56,7 +56,9 @@ def evaluate(model, matrix, split, live, save_path):
             {
                 "prc": [
                     {"precision": p, "recall": r, "threshold": t}
-                    for p, r, t in prc_points
+                    # https://github.com/scikit-learn/scikit-learn/pull/26194/files
+                    # Threshold can be inf now and Vega CLI tools don't like it.
+                    for p, r, t in prc_points if math.isfinite(t)
                 ]
             },
             fd,


shcheklein:
Doesn't fix it actually. The reason is in `ROC`, not in `PRC` which we don't generate manually atm :(.

PR #228, opened 2023-07-26
feat(example-get-started): add azure private remote

shcheklein:
As part of the https://github.com/iterative/studio-support/issues/89 . It add a way to use Azure private storage to generate a fixture like this: https://github.com/iterative/example-get-started-azure-private-fixture to then be used in the tests, or degug.

### DIFF
--- a/example-get-started/generate.sh
+++ b/example-get-started/generate.sh
@@ -18,7 +18,9 @@ OPT_INIT_DVC='true' # Default true.
 OPT_NON_DVC='false' # Default false.
 OPT_BRANCHES='true' # Default true.
 OPT_TAGS='true' # Default true.
-OPT_REMOTE='public-s3' # Default "public-s3". Other options: "public-s3", "private-http", "private-ssh", etc.
+# Default "public-s3". Other options: "public-s3", "private-http", "private-ssh", etc.
+# See the details below in the `init_remote_storage` and in the README.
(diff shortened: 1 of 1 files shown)


daavoo (approved):

example-get-started/generate.sh
```diff
@@ -76,6 +77,14 @@ init_remote_storage() {
     dvc remote add -f -d $OPT_REMOTE ssh://dvc@35.194.53.251/home/dvc/storage
     dvc remote modify $OPT_REMOTE keyfile /Users/ivan/.ssh/dvc_gcp_remotes_rsa
   fi
+
+  if [ $OPT_REMOTE == 'private-azure' ]; then
+    # Make sure that you have connection string in your env or some other way
+    # provide credentials for the `dvcprivate` storage account. Copy the connection
+    # string from the Azure portal and export it with
+    # `AZURE_STORAGE_CONNECTION_STRING`
+    dvc remote add -f -d $OPT_REMOTE azure://nlp
```
daavoo, line 86: minor: should we use a more descriptive name for the bucket?

shcheklein: tbh, was thinking to migrate our example repos to something like `nlp`, `cv`, etc. But we can come up with some better names of course. For Azure - let's keep it as-is since I don't think it will become an official example or anything.

PR #229, opened 2023-07-28
Update to YOLO. Add Sagemaker deployment

daavoo:
## Migrate to YOLO

Make the decision because:

- It was the framework I was already using on a similar project 😅
- It feels like it is more popular these days compared to fast.ai.
Not only in general, but in our particular case more than 1 customer/prospect has explicitly reported using YOLO but (afaik) we did not hear that from fast.ai.
- The framework logic for training felt simpler
If we remove the custom callback, it is `YOLO.train` vs ~60 lines of fast.ai

## Sagemaker deployment

Not that I am sure the approach here is the best but at least we could start the discussion.
Create https://www.notion.so/iterative/Sagemaker-Deployment-fce94fa0d81c44f395c10fc32940ab82#f3773819ebce4659ad397ed92686d56c to discuss.

---

TODO:

- [ ] Update README
- [ ] Update dvc.org references to the repo


shcheklein:
@daavoo could you point me please to some details / summary or update the description please - e.g. why YOLO (do we want / do we use the callback there?), what else has changed any why, etc. That make it easier for me to review. Thanks!

example-get-started-experiments/code/src/train.py
```diff
@@ -1,81 +1,85 @@
...
-            type="model",
-            name="pool-segmentation",
-            desc="This is a Computer Vision (CV) model that's segmenting out swimming pools from satellite images.",
-            labels=["cv", "segmentation", "satellite-images", params.train.arch],
-        )
+from fire import Fire
+from ultralytics import YOLO
+
+
+def add_callbacks(live, yolo):
```
daavoo, line 9: @shcheklein I am **not** using the built-in callback because:

- I couldn't find a way to customize arguments (i.e. pass a custom dir, disable report)
It seems the other loggers in ultralytics rely on env vars.

- Most of the image plots didn't feel useful / were redundant with existing plots.
I was planning to send a patch to ultralytics repo to skip some of the images (i.e. the confusion matrix, a big image with all the linear plots, etc)

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,58 @@
...
+    - run: pip install -r requirements.txt
+
+    - run: dvc pull results/train/artifacts/best.pt.dvc
+
+    - env:
+        VERSION: ${{steps.clean_version_name.outputs.version}}
+      run: |
+        bash sagemaker/bundle_and_upload_model.sh \
+        results/train/artifacts/best.pt \
+        s3://pool-segmentation/$VERSION/model.tar.gz
```
daavoo, line 49: This is the `A` option documented here : https://www.notion.so/iterative/Sagemaker-Deployment-fce94fa0d81c44f395c10fc32940ab82#735578764d964f9d9c6d7ab9d8c5fa39

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    - env:
+        VERSION: ${{steps.clean_version_name.outputs.version}}
+      run: |
+        python sagemaker/deploy_model.py \
+        --name $VERSION \
+        --model_data s3://pool-segmentation/$VERSION/model.tar.gz \
+        --role ${{ secrets.AWS_ROLE_TO_ASSUME }} \
+        --instance_type 'ml.c4.xlarge'
```
daavoo, line 58: This is option  `A`  of  `How to create endpoints based on MR events`.

Explained here : https://www.notion.so/iterative/Sagemaker-Deployment-fce94fa0d81c44f395c10fc32940ab82#f3773819ebce4659ad397ed92686d56c

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,58 @@
+name: Deploy model
+
+on:
+  push:
+    # When a new version is registered in Studio Model Registry
+    tags:
```
daavoo, line 6: Did not use gto action because it did simplify / add much value for the deployment pattern used

example-get-started-experiments/code/sagemaker/deploy_model.py
```diff
@@ -0,0 +1,46 @@
+import logging
```
daavoo, line 1: This is just standard code suggested by sagemaker docs :

https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html#deploy-pytorch-models

example-get-started-experiments/code/src/create_yolo_dataset.py
daavoo, line 1: This file could be removed if we create a version of the dataset in YOLO format instead of the current format with png masks

example-get-started-experiments/code/src/inference.py
daavoo, line 1: Just a simple script to showcase how to query the endpoint

shcheklein:
@daavoo 

> Make the decision because:

thanks for clarifying this. does it make it easier to deploy? I wonder if those changes are related at all to each other? could have we done them one my one? (easier to review, etc)

daavoo:
> thanks for clarifying this. does it make it easier to deploy? 

I found a working example of YOLO + Sagemaker that I could adapt but I did not find that for fast.ai 

> I wonder if those changes are related at all to each other? could have we done them one my one? (easier to review, etc)

I can split into 2 parts. Yolo and then sagemaker

shcheklein:
> I found a working example of YOLO + Sagemaker that I could adapt but I did not find that for fast.ai

make sense. Nw, let's keep it as is.

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,58 @@
...
+    - uses: aws-actions/configure-aws-credentials@v2
+      with:
+        aws-region: us-east-1
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.8'
```
shcheklein, line 36: doesn' work with the newer one?

example-get-started-experiments/code/.github/workflows/deploy-model.yml
shcheklein: can we avoid moving and rebundling the model?

example-get-started-experiments/code/README.md
```diff
@@ -1,7 +1,4 @@
-[![DVC](https://img.shields.io/badge/-Open_in_Studio-grey.svg?style=flat-square&logo=dvc)](https://studio.iterative.ai/team/Iterative/projects/example-get-started-experiments-y8toqd433r) 
-[![DVC-metrics](https://img.shields.io/badge/dynamic/json?style=flat-square&colorA=grey&colorB=F46737&label=Dice%20Metric&url=https://github.com/iterative/example-get-started-experiments/raw/main/results/evaluate/metrics.json&query=dice_multi)](https://github.com/iterative/example-get-started-experiments/raw/main/results/evaluate/metrics.json)
```
shcheklein, line 2: why removing these?

example-get-started-experiments/code/README.md
```diff
@@ -11,8 +8,6 @@ This is an auto-generated repository for use in [DVC](https://dvc.org)
 This is a Computer Vision (CV) project that solves the problem of segmenting out 
 swimming pools from satellite images. 
 
-[Example results](./results/evaluate/plots/images/)
```
shcheklein, line 14: should we still try to include some examples?

example-get-started-experiments/code/data/.gitignore
```diff
@@ -1,3 +1,2 @@
 /pool_data
-/test_data
-/train_data
+/yolo_dataset
```
shcheklein, line 2: does it need to be prefixed with `yolo_` - is that dataset yolo specific?

example-get-started-experiments/code/src/inference.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def inference(
+    img_path: str,
+    endpoint_name: Optional[str] = None,
+    model_path: Optional[str] = None,
+    output_path: str = "predictions",
+):
+
+    if endpoint_name is not None and model_path is not None:
```
shcheklein, line 19: why do we need two inferences (two python scripts) ... both of them something related to endpoints

example-get-started-experiments/code/src/train.py
shcheklein: This code below looks complicated for the example repo . It should not belong there I think. 

> Most of the image plots didn't feel useful / were redundant with existing plots.
I was planning to send a patch to ultralytics repo to skip some of the images (i.e. the confusion matrix, a big image with all the linear plots, etc)

I would let users decide on that in the UX / UI (that should handle that case)

> I couldn't find a way to customize arguments (i.e. pass a custom dir, disable report)
It seems the other loggers in ultralytics rely on env vars.

If we go with YOLO I think we don't have other options but figure this out - we can't maintain that level of duplication + code below is not simple at all to my taste (fine for an internal callback, but not for the example repo, especially not if we need to put it into notebooks)

example-get-started-experiments/code/src/train.py
```diff
@@ -1,81 +1,85 @@
...
+        os.remove("data/yolo_dataset/val.cache")
+    except FileNotFoundError:
+        pass
+    shutil.rmtree("./runs", ignore_errors=True)
+    shutil.rmtree("./weights", ignore_errors=True)
 
 
 if __name__ == "__main__":
-    train()
+    Fire(train)
```
shcheklein, line 85: newline

example-get-started-experiments/generate.sh
```diff
@@ -75,17 +76,17 @@ dvc pull
...
 
 cp -r $HERE/code/notebooks .
+cp -r $HERE/code/sagemaker .
 git add .
 git commit -m "Add notebook using DVCLive"
 
+sudo apt-get update && sudo apt-get install ffmpeg libsm6 libxext6 -y
 pip install -r requirements.txt --extra-index-url https://download.pytorch.org/whl/cu118
 pip install jupyter
+yolo settings datasets_dir=/workspaces/example-repos-dev/example-get-started-experiments/build/example-get-started-experiments/data
```
shcheklein, line 86: so, how will it look like in docs - we'll we have to explain all this in the get started?

shcheklein (commented):
May be it's worth splitting this indeed to unblock it faster. I like that we are trying yolo, it make  it more useful for actual customers and scenarios, but it seems we need to do some homework for that to look decent / simple to use and something that can fit into get started docs for the experiments. I would love to hear @dberenbaum opinion as well. May be we should do a leap of faith and just push though it and learn from it and keep doing improvements, etc. Let's decide on this. Duplication and complexity (e.g. custom logger) bothers me in this case.

On the deployment:

- I think we need an app that we could drop an image, so that we can demo it
- Do we keep instances / deployment running 24/7 for the demo purposes? What will be the workflow there? Can it be deployed in a serverless / on-demand mode?
- Good questions on the lifecycle - which endpoint we keep alive, etc. Don't know the answer yet to this.
- What is the lifecycle for code + models - can I change code and deploy with an older version of a model - how is it determined?

example-get-started-experiments/code/src/inference.py
daavoo: (`sagemaker/code/inference.py`) is bundled alongside the model file and uploaded to Sagemaker, it can't be used from a user perspective. 
It includes methods following conventions required by Sagemaker.

example-get-started-experiments/generate.sh
daavoo: No, this is only needed during generation inside codespaces because YOLO resolves the paths wrong inside `generate.sh`.
For someone cloning the repo it works without running the command

example-get-started-experiments/code/requirements.txt
```diff
@@ -1,4 +1,6 @@
 dvc[s3]>=3.0
 dvclive>=2.11.3
-fastai
-python-box
\ No newline at end of file
+fire
```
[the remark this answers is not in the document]
daavoo: fire is to remove the boilerplate of adding argparse. takes function signature and makes it work from CLI arguments
shapely needed to convert the png masks to yolo format, which expect polygons in a .txt

example-get-started-experiments/code/data/.gitignore
daavoo: is not needed. just to give a name that indicates the format

daavoo:
> May be it's worth splitting this indeed to unblock it faster. I like that we are trying yolo, it make it more useful for actual customers and scenarios, but it seems we need to do some homework for that to look decent / simple to use and something that can fit into get started docs for the experiments. I would love to hear @dberenbaum opinion as well. May be we should do a leap of faith and just push though it and learn from it and keep doing improvements, etc. Let's decide on this. Duplication and complexity (e.g. custom logger) bothers me in this case.

Moved the YOLO part to https://github.com/iterative/example-repos-dev/pull/232 .
Tried to simplify (removed custom callback; sent patch upstream).

Issue #230, opened 2023-08-02
Unclear reference to a pull request in example-get-started-experiments README.md


tibor-mach:
[Here](https://github.com/iterative/example-repos-dev/blob/af382934c4bc841463ba085b4387826205ca7b7a/example-get-started-experiments/code/README.md?plain=1#L111C1-L115C114) we have a reference in the README.md to a pull request which adds a model to the studio model registry. But there are no pull requests in that link, it just is just a link to the github pull requests page.

I am not sure if @aguschin meant that if you run the `dvc-studio` GH Actions workflow that you will get a pull request there or if it was supposed to actually refer to a specific pull request. Either way, I think this might be confusing to people who want to try things out.

I can update it to make it clearer, but please @aguschin let me know what the meaning was. I guess it was the first option, but I don't want to mess it up without asking. Thanks!

dberenbaum:
There's discussion about it in #204, but it looks like it should have been dropped as part of #207. @tibor-mach Feel free to submit a PR to drop it.

dberenbaum:
Fixed per #233

PR #232, opened 2023-08-03
Update to yolo

daavoo:
Deployed:

https://github.com/daavoo/example-get-started-experiments
https://studio.iterative.ai/team/Iterative/projects/example-get-started-experiments-bhm5eg8le6

---

Make the decision because:

- It was the framework I was already using on a similar project 😅
- It feels like it is more popular these days compared to fast.ai.
Not only in general, but in our particular case more than 1 customer/prospect has explicitly reported using YOLO but (afaik) we did not hear that from fast.ai.
- The framework logic for training felt simpler
See train.py vs previous with fast.ai
No need to separate evaluate.py stage because YOLO loads the best model and evaluates it at the end of training.

---

TODO: 

- [ ]  Update https://dvc.org/doc/start/experiments
- [ ] https://github.com/ultralytics/ultralytics/pull/4143


example-get-started-experiments/code/README.md
```diff
@@ -107,47 +103,3 @@ This tag also contains a GitHub Actions workflow that reruns the pipeline if any
  changes are introduced to the pipeline-related files. 
 [CML](https://cml.dev/) is used in this workflow to provision a cloud-based GPU 
 machine as well as report model performance results in Pull Requests.
-
-## Deploying the model
```
daavoo, line 111: Removing this section for now.
Will properly close https://github.com/iterative/example-repos-dev/issues/230 on a subsequent P.R. adding the sagemaker code

dberenbaum:
Thoughts on yolo:
* I like that it's a more popular, high-level framework.
* Am I reading it wrong, or are the predictions pretty bad?
* Not sure if the simplicity is a plus here. It is easy to read, but it's maybe too trivial? Do I really need to modularize my training code if it's 2 lines?

A couple small Studio UI issues:

* Did you manually hide all older commits in Studio? I only see the latest commit and its experiments.
* Similarly, it looks like lots of metrics are hidden. If we want to use a more realistic example in yolo, I think it would be better to show more metrics.

example-get-started-experiments/code/requirements.txt
```diff
@@ -1,4 +1,5 @@
 dvc[s3]>=3.0
 dvclive>=2.11.3
-fastai
-python-box
\ No newline at end of file
+fire
```
dberenbaum, line 3: Minor, but I vote to drop fire and use argparse. The scripts will still be plenty short and I think they will be more transparent.

example-get-started-experiments/code/src/train.py
```diff
@@ -1,81 +1,12 @@
...
-        models_dir.mkdir(exist_ok=True)
-        learn.export(fname=(models_dir / "model.pkl").absolute())
-        live.log_artifact(
-            str(models_dir / "model.pkl"),
-            type="model",
-            name="pool-segmentation",
-            desc="This is a Computer Vision (CV) model that's segmenting out swimming pools from satellite images.",
-            labels=["cv", "segmentation", "satellite-images", params.train.arch],
-        )
+    yolo.train(data="datasets/yolo_dataset.yaml", epochs=epochs, imgsz=imgsz, **kwargs)
```
dberenbaum, line 8: Why not track `datasets/yolo_dataset.yaml` in `dvc.yaml`?

example-get-started-experiments/generate.sh
```diff
@@ -67,56 +71,45 @@ tick
...
 pip install jupyter
-jupyter nbconvert --execute 'notebooks/TrainSegModel.ipynb' --inplace
-# Apply best experiment
-BEST_EXP_ROW=$(dvc exp show --drop '.*' --keep 'Experiment|evaluate/dice_multi|base_lr' --csv --sort-by evaluate/dice_multi | tail -n 1)
-BEST_EXP_NAME=$(echo $BEST_EXP_ROW | cut -d, -f 1)
-BEST_EXP_BASE_LR=$(echo $BEST_EXP_ROW | cut -d, -f 3)
-dvc exp apply $BEST_EXP_NAME
+yolo settings datasets_dir="/workspaces/example-repos-dev/example-get-started-experiments/build/example-get-started-experiments/datasets/"
+yolo settings runs_dir="/workspaces/example-repos-dev/example-get-started-experiments/build/example-get-started-experiments/runs/"
+yolo settings weights_dir="/workspaces/example-repos-dev/example-get-started-experiments/build/example-get-started-experiments/weights/"
```
dberenbaum, line 90: Why is it needed in this script but doesn't seem to be required when running in the published repo?

example-get-started-experiments/generate.sh
```diff
@@ -67,56 +71,45 @@ tick
...
+jupyter nbconvert --execute 'TrainSegModel.ipynb' --inplace
 git add .
 tick
 git commit -m "Run notebook and apply best experiment"
```
dberenbaum, line 94: Is it still applying the best experiment?

example-get-started-experiments/generate.sh
```diff
@@ -67,56 +71,45 @@ tick
...
-  -d src/train.py -d data/train_data \
-  -o models/model.pkl \
-  python src/train.py
-
-dvc stage add -n evaluate \
-  -p base,evaluate \
-  -d src/evaluate.py -d models/model.pkl -d data/test_data \
-  python src/evaluate.py
+  -d src/train.py -d datasets/yolo_dataset/train -d datasets/yolo_dataset/val \
+  "python src/train.py \${train}"
```
dberenbaum, line 110: Feels odd to not have the model as an output of the stage here to me.

Also, why do have train and val datasets as separate deps/outs instead of tracking the whole dataset as one?

dberenbaum:
Overall, I think it's great and shows off how powerful and simple the tools can be, but feels more like a showcase of functionality than a way to teach about dvc experiments. I would like to at least take some steps on https://github.com/iterative/dvclive/issues/603 if we make this our default example project. Although it's hard to maintain them all as auto-generated repos, I think we need many examples of different frameworks, so I would urge that we should try to keep both projects in some form instead of replacing one with the other.

example-get-started-experiments/generate.sh
daavoo: Because in the codespace we are running `bash generate.sh` and inside that we are changing the working directory which messes with where YOLO is looking for the dataset

example-get-started-experiments/generate.sh
daavoo: No, do you think we should keep that thing? I am not even sure if people realize what was going one without checking this script

example-get-started-experiments/generate.sh
daavoo: > Feels odd to not have the model as an output of the stage here to me.

I can make the changes for that. I know odd but couldn't find a reason why it matters (maybe I should add additional stage)

> Also, why do have train and val datasets as separate deps/outs instead of tracking the whole dataset as one?

So in Studio you can use Data columns to show the number of files separately.

daavoo:
> * Am I reading it wrong, or are the predictions pretty bad?

Not sure how/what you are reading. I could talk for a while about metrics for these tasks 😅

The metric currently selected is an instance segmentation one (pmAP 0.5-0.95) and it doesn't hold the same meaning as accuracy or similar (like, I expect `0.9` to consider it good)

> * Not sure if the simplicity is a plus here. It is easy to read, but it's maybe too trivial? Do I really need to modularize my training code if it's 2 lines?

The point is not really to modularize but to use `exp run` and its associated features, right?

> * How is dvclive even invoked here? Any idea why https://docs.ultralytics.com/reference/yolo/utils/callbacks/dvc/ is giving a 404?

It is automatically added as a callback if it is installed: 

https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/callbacks/dvc.py

This is the way YOLO currently handles the integrations.

> * In general, I worry there's so much magic happening in both dvc and yolo here that it's hard for the user to learn anything from it or generalize to their use case.

So, I am assuming you did not have that worry with the previous fast.ai code because the callback was passed explicitly?
I remember the feedback was that it was too complicated because it used the context manager, etc.

> A couple small Studio UI issues:
> 
> * Did you manually hide all older commits in Studio? I only see the latest commit and its experiments.

Yes, feel free to add/show whatever.

> * Similarly, it looks like lots of metrics are hidden. If we want to use a more realistic example in yolo, I think it would be better to show more metrics.

Same as above, just the view I added, doesn't have to be what we show. It is what I prefer but I don't know if that is the best way to show as an example.
I do worry about the fallacy of making things look "complex" for the sake of appearance.

daavoo:
>  but feels more like a showcase of functionality than a way to teach about dvc experiments

Sorry, maybe it was already implicit in the previous comments but could you summarize the point that makes it the case for this example vs the current one? 

What I got as main differences:

- no evaluate stage
- train code to simple

Is there something else?

example-get-started-experiments/code/src/train.py
daavoo: No reason. Can totally do it. 
It is just a description of where paths are because YOLO wants it to be in a file so felt like not really relevant

example-get-started-experiments/code/requirements.txt
daavoo: will do

example-get-started-experiments/generate.sh
dberenbaum: No, I just meant we should change the commit message

example-get-started-experiments/generate.sh
dberenbaum: > I can make the changes for that. I know odd but couldn't find a reason why it matters (maybe I should add additional stage)

Does https://github.com/iterative/dvclive/pull/631 break anything here since it skips caching if `_inside_dvc_exp=True`?

daavoo: > Does https://github.com/iterative/dvclive/pull/631 break anything here since it skips caching if _inside_dvc_exp=True?

Because the .dvc file is created at notebook stage, I believe there is no difference because DVC will (dvc) commit the changes at the end

dberenbaum:
> The metric currently selected is an instance segmentation one (pmAP 0.5-0.95) and it doesn't hold the same meaning as accuracy or similar (like, I expect `0.9` to consider it good).

Sorry, I was reading it wrong 😄 . I have used mAP but here I was looking at the image masks. I think it is just harder to see whether the red bounding boxes overlay an actual pool. Anyway, disregard that comment.

> Sorry, maybe it was already implicit in the previous comments but could you summarize the point that makes it the case for this example vs the current one?

I don't have that strong an opinion on which one should be the "default." I think we need to keep both (and in that case, I'm fine with keeping the deployment in this one) because they showcase different features and frameworks and we need more examples.

Reasons why I wouldn't push to make this one the "default":

1. Docs need to be reworked and I don't see that it would be worth the time.
2. If I'm not using yolo, it doesn't show me anything about how to use dvclive. By default, I would rather have an example that invokes `Live()` or an explicit `DVCLiveCallback()`.

daavoo:
> Sorry, I was reading it wrong 😄 . I have used mAP but here I was looking at the image masks. I think it is just harder to see whether the red bounding boxes overlay an actual pool. Anyway, disregard that comment.

Yes, the default built-in visualization is ... not great.

> I don't have that strong an opinion on which one should be the "default." I think we need to keep both (and in that case, I'm fine with keeping the deployment in this one) because they showcase different features and frameworks and we need more examples.
> 
> Reasons why I wouldn't push to make this one the "default":
> 
> 1. Docs need to be reworked and I don't see that it would be worth the time.
> 2. If I'm not using yolo, it doesn't show me anything about how to use dvclive. By default, I would rather have an example that invokes `Live()` or an explicit `DVCLiveCallback()`.

Makes sense.
For the record about 2, at least HuggingFace appears to now be also defaulting to automatically include the callback if the library is installed, so assuming we start moving the callbacks to their repos, more examples might start to look like this

daavoo:
Will make it another repo, converting to draft for now

PR #233, opened 2023-08-10
222 add deployment of example get started experiments model

daavoo:
Add Sagemaker deployment.

https://github.com/iterative/example-get-started-experiments/actions/workflows/deploy-model.yml

https://us-east-2.console.aws.amazon.com/sagemaker/home?region=us-east-2#/endpoints/results-train-pool-segmentation-v0-1-0-dev

```bash
export AWS_DEFAULT_REGION=us-east-2
python src/endpoint_predict.py \
--img_path data/test_data/REGION_1-24_0_1024_0_1024.jpg \
--endpoint_name results-train-pool-segmentation-v0-1-0-dev
```


example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - id: clean_version_name
+      run: echo "version=$(echo ${{ github.ref_name || github.event.inputs.version }} | sed 's/[^a-zA-Z0-9-]/-/g')" >> "$GITHUB_OUTPUT"
+
+    - uses: aws-actions/configure-aws-credentials@v2
+      with:
+        aws-region: us-east-2
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
```
daavoo, line 37: I need some help from the @iterative/platform team (or someone more AWS savvy than me).

When I use:

```
      - uses: aws-actions/configure-aws-credentials@v1
        with:
          aws-region: us-east-2
          role-to-assume: arn:aws:iam::342840881361:role/SandboxUser
          role-duration-seconds: 43200
 ```
 
The sagemaker API fails with:


```
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the CreateModel operation: Could not access model data at s3://dvc-public/remote/get-started-pools/files/md5/8c/1d4e43ef621114f9798f9301e98cb3. Please ensure that the role "arn:aws:iam::***:role/SandboxUser" exists and that its trust relationship policy allows the action "sts:AssumeRole" for the service principal "sagemaker.amazonaws.com". Also ensure that the role has "s3:GetObject" permissions and that the object is located in us-east-2. If your Model uses multiple models or uncompressed models, please ensure that the role has "s3:ListBucket" permission.
```

Comment (jesper7) [reply example-get-started-experiments/code/.github/workflows/deploy-model.yml:37]: Responded in private

File: example-get-started-experiments/code/.github/workflows/deploy-model.yml [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - id: clean_version_name
+      run: echo "version=$(echo ${{ github.ref_name || github.event.inputs.version }} | sed 's/[^a-zA-Z0-9-]/-/g')" >> "$GITHUB_OUTPUT"
+
+    - uses: aws-actions/configure-aws-credentials@v1
```
tapadipti, line 32:
```suggestion
    - uses: aws-actions/configure-aws-credentials@v2
```

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,57 @@
...
+      with:
+        aws-region: us-east-2
+        role-to-assume: arn:aws:iam::342840881361:role/SandboxUser
```
tapadipti, line 35: Should we make this a GH secret?

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,57 @@
...
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.8'
+        cache: 'pip'
+        cache-dependency-path: requirements.txt
+
+    - run: pip install -r requirements.txt
+
+    - run:  dvc remote add -d --local storage s3://dvc-public/remote/get-started-pools
+
```
tapadipti, line 48:
```suggestion
    - name: "Use GTO"
      id: gto
      uses: iterative/gto-action@14723404a00bb0c1e759c02ffcd24279df5815c2
      
    - name: "Get model url"
      if: steps.gto.outputs.event == 'assignment'
      run: |
        echo "MODEL_URL=$(dvc get --show-url --rev ${{ github.sha }} https://github.com/${{ github.repository }} [Add the path to the model file, eg. output/mymodel.tar.gz])" >> "$GITHUB_ENV"
        
    - name: "Run deployment python script"
      if: steps.gto.outputs.event == 'assignment'
      run: |
        python sagemaker/deploy_model.py ${{ steps.gto.outputs.version }} ${{ steps.gto.outputs.stage }} ${{ env.MODEL_URL }}
```

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,57 @@
...
+    - env:
```
tapadipti, line 49: @daavoo I've added suggestion about using the GTO action to find the event, version and stage and use it to create the deployment, so that there can be different models/configs per version/stage and different endpoints per stage.

tapadipti (commented):
Added some change suggestions for the deploy GH action. Will add related changes to the deploy-model file.

example-get-started-experiments/code/sagemaker/deploy_model.py
```diff
@@ -0,0 +1,54 @@
...
+            "SAGEMAKER_MODEL_SERVER_TIMEOUT": "3600",
+            "TS_MAX_RESPONSE_SIZE": "2000000000",
+            "TS_MAX_REQUEST_SIZE": "2000000000",
+            "MMS_MAX_RESPONSE_SIZE": "2000000000",
+            "MMS_MAX_REQUEST_SIZE": "2000000000",
+        },
+    )
+
+
+    return model.deploy(
```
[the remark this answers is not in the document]
daavoo: There is no need to use that code, the `deploy` method accepts `serverless_inference_config` as an argument

example-get-started-experiments/code/.github/workflows/deploy-model.yml
daavoo: I updated the workflow to parse the GTO in a separate job, so we don't do any unnecessary steps in case it is not an assignment event

example-get-started-experiments/code/sagemaker/deploy_model.py
```diff
@@ -0,0 +1,78 @@
...
+    version: str,
+    model_data: str,
+    role: str,
+):
+    sagemaker_logger = logging.getLogger("sagemaker")
+    sagemaker_logger.setLevel(logging.DEBUG)
+    sagemaker_logger.addHandler(logging.StreamHandler(sys.stdout))
+
+    composed_name =  re.sub(
+        r"[^a-zA-Z0-9\-]", "-", f"{name}-{version}-{stage}")
```
dberenbaum, line 36: Do you know what possible characters can appear in `f"{name}-{version}-{stage}"` that would need to be replaced here? I guess it's related to https://github.com/iterative/dvc/issues/9821?

example-get-started-experiments/code/src/train.py
```diff
@@ -68,6 +68,7 @@ def train():
         models_dir = Path("models")
         models_dir.mkdir(exist_ok=True)
         learn.export(fname=(models_dir / "model.pkl").absolute())
```
dberenbaum, line 70: Is this still needed/useful? Seems odd to have `export` and `save` here.

example-get-started-experiments/generate.sh
```diff
@@ -109,14 +108,18 @@ dvc remove models/model.pkl.dvc
 dvc stage add -n train \
   -p base,train \
   -d src/train.py -d data/train_data \
-  -o models/model.pkl \
+  -o models/model.pkl -o models/model.pth \
```
dberenbaum, line 111: Related to the comment above, do we still need `model/model.pkl`?

example-get-started-experiments/code/src/train.py
daavoo: I would have to change the evaluate stage to use the `.pth` format. 
I was not able to make sagemaker work with the fast.ai exported in `.pkl` format.

example-get-started-experiments/code/sagemaker/deploy_model.py
daavoo: > I guess it's related to [iterative/dvc#9821](https://github.com/iterative/dvc/issues/9821)?

This regex is only relevant for Sagemaker and it is enforced by the API

example-get-started-experiments/generate.sh
```diff
@@ -109,14 +108,18 @@ dvc remove models/model.pkl.dvc
...
   python src/train.py
 
 dvc stage add -n evaluate \
   -p base,evaluate \
   -d src/evaluate.py -d models/model.pkl -d data/test_data \
   python src/evaluate.py
 
+dvc stage add -n sagemaker \
+  -d models/model.pth -o model.tar.gz \
+  'cp models/model.pth sagemaker/code/model.pth && cd sagemaker && tar -cpzf model.tar.gz code/ && cd .. && mv sagemaker/model.tar.gz .  && rm sagemaker/code/model.pth'
```
dberenbaum, line 121: Some minor thoughts to try to clean this up a bit:

I haven't tested this, but wondering if we can simplify this at all with something like this?

```suggestion
  'cp models/model.pth sagemaker/out && cp sagemaker/code sagemaker/out && tar -cpzf model.tar.gz sagemaker/out'
```

Also, I wonder if it would be better to append directly to `dvc.yaml` and use list syntax for `cmd` instead of `&&`?

example-get-started-experiments/generate.sh
```diff
@@ -90,11 +90,10 @@ git add .
 tick
 git commit -m "Run notebook and apply best experiment"
 git tag -a "1-notebook-dvclive" -m "Experiment using Notebook"
-gto register results/train:pool-segmentation --version v1.0.0
-gto assign results/train:pool-segmentation --version v1.0.0 --stage dev
 
 
 cp -r $HERE/code/src .
+cp -r $HERE/code/sagemaker .
```
dberenbaum, line 96: Could we add all this later and make it a separate tag?

dberenbaum:
I don't seem to have access:

Screenshot 2023-08-14 at 1 36 08 PM

```
$ python src/endpoint_prediction.py \
--img_path data/test_data/REGION_1-24_0_1024_0_1024.jpg \
--endpoint_name results-train-pool-segmentation-v0-1-0-dev
Traceback (most recent call last):
  File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 53, in <module>
    endpoint_prediction(args.img_path, args.endpoint_name, args.output_path)
  File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 32, in endpoint_prediction
    result = predictor.predict(img_bytes)[0]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/sagemaker/base_predictor.py", line 185, in predict
    response = self.sagemaker_session.sagemaker_runtime_client.invoke_endpoint(**request_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 964, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ExpiredTokenException) when calling the InvokeEndpoint operation: The security token included in the request is expired
(example-get-started-experiments) dave@davids-air:~/Code/example-get-started-experiments [main] 13:32:25
$ ~/sts.sh 363718
Configuring AWS with token 363718
(example-get-started-experiments) dave@davids-air:~/Code/example-get-started-experiments [main] 13:32:36
$ python src/endpoint_prediction.py \
--img_path data/test_data/REGION_1-24_0_1024_0_1024.jpg \
--endpoint_name results-train-pool-segmentation-v0-1-0-dev
Traceback (most recent call last):
  File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 53, in <module>
    endpoint_prediction(args.img_path, args.endpoint_name, args.output_path)
  File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 32, in endpoint_prediction
    result = predictor.predict(img_bytes)[0]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/sagemaker/base_predictor.py", line 185, in predict
    response = self.sagemaker_session.sagemaker_runtime_client.invoke_endpoint(**request_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 964, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the InvokeEndpoint operation: User: arn:aws:iam::260760892802:user/dave is not authorized to perform: sagemaker:InvokeEndpoint on resource: arn:aws:sagemaker:us-east-2:260760892802:endpoint/results-train-pool-segmentation-v0-1-0-dev because no identity-based policy allows the sagemaker:InvokeEndpoint action
```

What is our plan here? Not a blocker, but do we want to work towards making it public?

dberenbaum (approved):
It would be nice to address my comments from this PR, but I don't see any actual blockers here. Nice work @daavoo!

example-get-started-experiments/code/sagemaker/deploy_model.py
dberenbaum: I think the rationale for the naming restrictions in gto were to make it more likely they would be compatible with systems like sagemaker. Not sure it really helps though. We still hit these restrictions and users don't know why they can't use these characters.

example-get-started-experiments/code/src/train.py
daavoo: > Have you tried it? I would say let's do it if it's not that hard.

I assume it is possible but I didn't try to not include other changes

example-get-started-experiments/generate.sh
daavoo: > Also, I wonder if it would be better to append directly to dvc.yaml and use list syntax for cmd instead of &&?

I didn't manage to use list cmd with `stage add` (which we use in the `generate.sh` script , but might be missing something

daavoo:
> I don't seem to have access:

Can you try with the Sandbox account? 
Also, make sure you set `us-east-2` as AWS region when querying

> What is our plan here? Not a blocker, but do we want to work towards making it public?

I assume we don't want to make the actual endpoint public, but rather a simple UI that queries the endpoint.
I was assuming that, for now, we would be using it for live demos and using the sandbox account.

dberenbaum:
> Can you try with the Sandbox account?
> Also, make sure you set `us-east-2` as AWS region when querying

Thanks, that helped, but now I'm getting a timeout error:

```
$ python src/endpoint_prediction.py \
--img_path data/test_data/REGION_1-24_0_1024_0_1024.jpg \
--endpoint_name results-train-pool-segmentation-v0-1-0-dev

Traceback (most recent call last):
  File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 53, in <module>
    endpoint_prediction(args.img_path, args.endpoint_name, args.output_path)
  File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 32, in endpoint_prediction
    result = predictor.predict(img_bytes)[0]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/sagemaker/base_predictor.py", line 185, in predict
    response = self.sagemaker_session.sagemaker_runtime_client.invoke_endpoint(**request_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 964, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ModelError: An error occurred (ModelError) when calling the InvokeEndpoint operation: Received server error (0) from model with message "Your invocation timed out while waiting for a response from model container. Review the latency metrics in Amazon CloudWatch, resolve the issue, and try again.". See https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logEventViewer:group=/aws/sagemaker/Endpoints/results-train-pool-segmentation-v0-1-0-dev in account 342840881361 for more information.
```

I also see errors in the logs in https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logsV2:log-groups/log-group/$252Faws$252Fsagemaker$252FEndpoints$252Fresults-train-pool-segmentation-v0-1-0-dev/log-events/AllTraffic$252F44267aee8024d8ef1612febe258e9378-08a54f8ef3504be4b8e6e736d1e78a67.

daavoo:
> > Can you try with the Sandbox account?
> > Also, make sure you set `us-east-2` as AWS region when querying
> 
> Thanks, that helped, but now I'm getting a timeout error:
> 
> ```
> $ python src/endpoint_prediction.py \
> --img_path data/test_data/REGION_1-24_0_1024_0_1024.jpg \
> --endpoint_name results-train-pool-segmentation-v0-1-0-dev
> 
> Traceback (most recent call last):
>   File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 53, in <module>
>     endpoint_prediction(args.img_path, args.endpoint_name, args.output_path)
>   File "/Users/dave/Code/example-get-started-experiments/src/endpoint_prediction.py", line 32, in endpoint_prediction
>     result = predictor.predict(img_bytes)[0]
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/sagemaker/base_predictor.py", line 185, in predict
>     response = self.sagemaker_session.sagemaker_runtime_client.invoke_endpoint(**request_args)
>                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 530, in _api_call
>     return self._make_api_call(operation_name, kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/dave/micromamba/envs/example-get-started-experiments/lib/python3.11/site-packages/botocore/client.py", line 964, in _make_api_call
>     raise error_class(parsed_response, operation_name)
> botocore.errorfactory.ModelError: An error occurred (ModelError) when calling the InvokeEndpoint operation: Received server error (0) from model with message "Your invocation timed out while waiting for a response from model container. Review the latency metrics in Amazon CloudWatch, resolve the issue, and try again.". See https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logEventViewer:group=/aws/sagemaker/Endpoints/results-train-pool-segmentation-v0-1-0-dev in account 342840881361 for more information.
> ```
> 
> I also see errors in the logs in https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logsV2:log-groups/log-group/$252Faws$252Fsagemaker$252FEndpoints$252Fresults-train-pool-segmentation-v0-1-0-dev/log-events/AllTraffic$252F44267aee8024d8ef1612febe258e9378-08a54f8ef3504be4b8e6e736d1e78a67.

I will take a look tomorrow.
I tested it with a different instance type and I assume that the current serverless configuration is too small

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,70 @@
...
+    - run:  dvc remote add -d --local storage s3://dvc-public/remote/get-started-pools
+
+    - run: |
+        MODEL_DATA=$(dvc get --show-url . model.tar.gz)
+        python sagemaker/deploy_model.py \
+        --name ${{ needs.parse.outputs.name }} \
+        --stage ${{ needs.parse.outputs.stage }} \
+        --version ${{ needs.parse.outputs.version }} \
+        --model_data $MODEL_DATA \
+        --role ${{ secrets.AWS_ROLE_TO_ASSUME }}
```
shcheklein, line 70: newline

example-get-started-experiments/code/.github/workflows/dvc-studio.yml
```diff
@@ -90,7 +90,7 @@ jobs:
       - uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-region: us-east-2
-          role-to-assume: arn:aws:iam::342840881361:role/SandboxUser
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
```
shcheklein, line 93: do we have to generalize it in this way? (an extra step for us to take care of)

shcheklein (commented):
Looks really great! looks simple and clen

example-get-started-experiments/code/.github/workflows/dvc-studio.yml
daavoo: It was suggested by @jesper7 and @tapadipti to use the role as a secret. I assume it has some security implications although I am not sure what is the danger of leaking a role name

daavoo:
> Since this uses one of our official demo repos, we could use this in [the blog post](https://docs.google.com/document/d/1N87_WttHBzxmwAoehu4HKpIrtIqEPM2UavZq82vo_hU/edit#heading=h.mjk189b4wmqq) instead of the demo-fashion-mnist that I have currently used. wdyt?

Makes sense to me. I would perhaps also use the opportunity to cut the scope of the post a little by dropping DVC details in favor of pointers to the dvc get-started pages

tapadipti:
> > Since this uses one of our official demo repos, we could use this in [the blog post](https://docs.google.com/document/d/1N87_WttHBzxmwAoehu4HKpIrtIqEPM2UavZq82vo_hU/edit#heading=h.mjk189b4wmqq) instead of the demo-fashion-mnist that I have currently used. wdyt?
>
> Makes sense to me. I would perhaps also use the opportunity to cut the scope of the post a little by dropping DVC details in favor of pointers to the dvc get-started pages

Ok. I'll share an updated version of the blog post tomorrow.
@shcheklein FYI since we were discussing this today morning.

daavoo:
Merging as the endpoint is now working. Don't hesitate to open followups

example-get-started-experiments/code/.github/workflows/deploy-model.yml
```diff
@@ -0,0 +1,70 @@
...
+        python-version: '3.8'
+        cache: 'pip'
+        cache-dependency-path: requirements.txt
+
+    - run: pip install -r requirements.txt
+
+    - run:  dvc remote add -d --local storage s3://dvc-public/remote/get-started-pools
+
+    - run: |
+        MODEL_DATA=$(dvc get --show-url . model.tar.gz)
```
tapadipti, line 64: @daavoo Doesn't this get the model data for the latest commit instead of the specific version we are trying to deploy? From the DVC docs, when `--rev` is not specified, `The latest commit (in the default branch) is used by default when this option is not specified.`

daavoo: I assume that the phrase in the docs is meant for the scenario where you pass a remote URL repo as the first argument. 

I think the behavior here is correct, but I am going to double-check.

I think it works because we are using the local repo `.` and so the current status of the workspace will be used. The workflow uses `actions/checkout` and runs on the git tag creation so the workspace will be the status of the revision where the tag was created

example-get-started-experiments/code/sagemaker/deploy_model.py
```diff
@@ -0,0 +1,78 @@
...
+    
+    model = PyTorchModel(
```
tapadipti, line 38: @daavoo If I created a SageMaker model for a given model version before deploying it in one environment (eg, `dev`), does this code recreate the same model before deploying it in another environment (eg, `prod`). Or does it skip recreating it and just return a reference to the old model?

example-get-started-experiments/code/sagemaker/deploy_model.py
```diff
@@ -0,0 +1,78 @@
...
+            "SAGEMAKER_MODEL_SERVER_TIMEOUT": "3600",
+            "TS_MAX_RESPONSE_SIZE": "2000000000",
+            "TS_MAX_REQUEST_SIZE": "2000000000",
+            "MMS_MAX_RESPONSE_SIZE": "2000000000",
+            "MMS_MAX_REQUEST_SIZE": "2000000000",
+        },
+    )
+
+
+    return model.deploy(
```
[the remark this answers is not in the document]
daavoo: > 

> @daavoo If I am trying to deploy a new model version to an existing stage, does this code create a new endpoint or does it update the existing endpoint? (I need to confirm this for the blog post)

Not sure I understand the first part but a new endpoint is created from combining the name, version, and stage, so any change to those 3 things result in a new endpoint

example-get-started-experiments/code/sagemaker/deploy_model.py
daavoo: It currently creates a new model. If we want to do what you said last, we should only add `stage` to the composed name after the model has been created.

example-get-started-experiments/code/src/train.py
```diff
@@ -68,6 +68,7 @@ def train():
...
+        torch.save(learn.model, (models_dir / "model.pth").absolute())
         live.log_artifact(
```
tapadipti, line 72: @daavoo If we logged the tar file of the model, then we wouldn't need to specify the file name `model.tar.gz` in the deployment script, right? The deployment script would be able to get the file name from the Git tag itself, which means that a single deployment script could be used for deploying several models with different names. Right?

example-get-started-experiments/code/sagemaker/deploy_model.py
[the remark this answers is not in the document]
tapadipti: So if I deploy v1 to prod today, and v2 to prod tomorrow, will I have 2 different endpoints with 2 different endpoint names? Shouldn't the prod endpoint always be the same? (just like studio prod version is always studio.iterative.ai) So that any clients running inference against the prod endpoint don't have to update their endpoint addresses each time there's a new deployment.

example-get-started-experiments/code/sagemaker/deploy_model.py
tapadipti: Referring to [this msg](https://iterativeai.slack.com/archives/C01SR9Q12LB/p1690971158412739?thread_ts=1690817210.212859&cid=C01SR9Q12LB), you said that the existing model would be re-used, right? Would removing `stage` from the model name make this happen?

example-get-started-experiments/code/sagemaker/deploy_model.py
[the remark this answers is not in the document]
tapadipti: The version shouldn't be part of the endpoint name, right?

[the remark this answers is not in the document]
daavoo: > Shouldn't the prod endpoint always be the same? 

I think it is a matter of opinion and depends on the pattern we expect/want to showcase:

A) The endpoint is directly queried by external consumers
B) There is some app&url that uses the right endpoint internally without exposing it to external consumers.

> (just like studio prod version is always studio.iterative.ai) So that any clients running inference against the prod endpoint don't have to update their endpoint addresses each time there's a new deployment.

That is the user-facing app&url and it only "points" to the correct software because there are gitops bumping the internal studio version (i.e. https://github.com/iterative/itops/pull/2238)

example-get-started-experiments/code/sagemaker/deploy_model.py
daavoo: > Would removing stage from the model name make this happen?

Yes

tapadipti: ok, I'll create a PR to remove stage from the model name.

example-get-started-experiments/code/sagemaker/deploy_model.py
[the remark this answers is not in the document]
tapadipti: So then we should also create a service to delete stale endpoints, right? Else, we will end up with a bunch of endpoints over time.

dberenbaum:
Agree with @tapadipti that it makes sense to have one endpoint per stage or per version. Otherwise, I think we kind of miss the point of the registry (you can deploy every update to a new endpoint without it). IMO one endpoint per stage makes the most sense to drive home the value of that field, and I think we should focus on this being a self-contained deployment (you can do deployment without needing a separate engineering team to pick up the new model endpoint).

example-get-started-experiments/code/sagemaker/deploy_model.py
tapadipti: [PR](https://github.com/iterative/example-repos-dev/pull/236)

example-get-started-experiments/code/sagemaker/deploy_model.py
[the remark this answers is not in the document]
daavoo: > So then we should also create a service to delete stale endpoints, right? Else, we will end up with a bunch of endpoints over time.

I would expect that to be configured / handled on the AWS side.

Anyhow, as per your suggestions, I will send a new P.R. creating/updating a single endpoint per stage

example-get-started-experiments/code/.github/workflows/dvc-studio.yml
dberenbaum: @daavoo How did you set it up for https://github.com/iterative/example-get-started-experiments? Should we add it to the readme here?

0x2b3bfa0: Note that roles can be public.

Issue #234, opened 2023-08-14
Use `live.log_image` for example-get-started plots


dberenbaum:
Per https://github.com/iterative/dvclive/issues/224#issuecomment-1271787193,

PR #235, opened 2023-08-15
example-get-started: update plots to use dvclive

dberenbaum:
Closes https://github.com/iterative/example-repos-dev/issues/234.

Updates two plots to use dvclive:
* precision recall (using `log_sklearn_plot` with additional kwarg `drop_intermediate=True`)
* feature importance (using `log_image` matplotlib support)

All the plots are now dvclive, which simplifies the code and `dvc.yaml` a bit. I ended up caching all of them, because I don't see much upside to keeping the tsv/json plots in git, and it simplifies `dvc.yaml`. We lose explaining how to make more custom plots, but I think that's outside the scope of get started. If we want more custom plots, we can try to think of more complex visualizations to include.

Todo for deployment:
- [ ] Deploy the updated code
- [ ] Deploy the updated repo (experiments, PRs, etc.; see readme)
- [ ] Update Studio demo 
- [ ] @shcheklein Do we need to update https://github.com/iterative/example-get-started-azure-private-fixture, https://github.com/iterative/example-get-started-http-private-fixture, and https://github.com/iterative/example-get-started-ssh-private-fixture?


example-get-started/code/src/evaluate.py
```diff
@@ -36,34 +36,19 @@ def evaluate(model, matrix, split, live, save_path):
     live.summary["avg_prec"][split] = avg_prec
     live.summary["roc_auc"][split] = roc_auc
 
-    # ... and plots...
+    # ... and plots
```
dberenbaum, line 39:
```suggestion
    # ... and plots...
```

example-get-started/code/src/evaluate.py
```diff
@@ -37,33 +37,18 @@ def evaluate(model, matrix, split, live, save_path):
...
     live.log_sklearn_plot("roc", labels, predictions, name=f"roc/{split}")
-
-    # ... but actually it can be done with dumping data points into a file:
-    # ROC has a drop_intermediate arg that reduces the number of points.
-    # https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_curve.html#sklearn.metrics.roc_curve.
-    # PRC lacks this arg, so we manually reduce to 1000 points as a rough estimate.
-    precision, recall, prc_thresholds = metrics.precision_recall_curve(
-        labels, predictions
+    # ... and precision recall plot...
+    # ... which passes `drop_intermediate=True` to the sklearn method
```
dberenbaum, line 43:
```suggestion
    # ... which passes `drop_intermediate=True` to the sklearn method...
```

dberenbaum:
Merging but not deploying yet because I am working on some larger changes and don't think it's worth it for this change alone.

PR #236, opened 2023-08-17
Remove stage from model name

tapadipti:
As discussed [here](https://github.com/iterative/example-repos-dev/pull/233#discussion_r1296920635)

### DIFF
--- a/example-get-started-experiments/code/sagemaker/deploy_model.py
+++ b/example-get-started-experiments/code/sagemaker/deploy_model.py
@@ -33,7 +33,7 @@ def deploy(
     sagemaker_logger.addHandler(logging.StreamHandler(sys.stdout))
 
     composed_name =  re.sub(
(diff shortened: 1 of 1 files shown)


dberenbaum:
Hi @tapadipti, per the linked discussion, WDYT of dropping the version instead of the stage?

daavoo:
> I'm not entirely sure if this change works, coz I haven't tried it. I made it based on what @daavoo [said](https://github.com/iterative/example-repos-dev/pull/233#discussion_r1297005704) would work. I meant to request your (and @shcheklein's) review in [a different PR](https://github.com/iterative/example-repos-dev/pull/237). I added it to this PR by mistake.

The P.R. doesn't currently work as expected because the `composed_name` is used also for create_endpoint.
2 different names should be built from the inputs.
I will update in a different P.R.

PR #237, opened 2023-08-21
Don't run dvc-studio.yml on tag push, coz otherwise MR version and st…

tapadipti:
…age assignment actions trigger this workflow

### DIFF
--- a/example-get-started-experiments/code/.github/workflows/dvc-studio.yml
+++ b/example-get-started-experiments/code/.github/workflows/dvc-studio.yml
@@ -3,6 +3,8 @@ name: DVC Studio Experiment
 on:
 
   push:
+    tags-ignore:
+      - '**'
 
   workflow_dispatch:
     inputs:


daavoo (approved):

dberenbaum:
Sorry, I think I misunderstood the workflow here. Aren't version and stage updates when the action should be triggered? Like in https://github.com/iterative/example-gto/blob/main/.github/workflows/gto-act-on-tags.yml?

daavoo:
> Sorry, I think I misunderstood the workflow here. Aren't version and stage updates when the action should be triggered? Like in https://github.com/iterative/example-gto/blob/main/.github/workflows/gto-act-on-tags.yml?

This change is for the workflow that runs an experiment

tapadipti:
> Aren't version and stage updates when the action should be triggered? 

We could trigger the action on both user actions - version registration and stage assignment - or just on stage assignment.

The convention we are following is that when a user assigns a stage, they intend to deploy the model to that stage. So, we create the endpoint during this user action. This step requires the SageMaker model to have been created.

The SageMaker model itself could be created at one of the following two points:
- we could create it when a users registers a version, coz they will likely assign a stage for that version, and so it's a good idea to keep the model created and ready.
- we could create it only after the user assigns a stage, coz until then there's no need to deploy the model, which means that the SageMaker model is not really needed until then.

Accordingly, we could trigger the GH action during version registration and stage assignment, or only during stage assignment.

Does this answer your question @dberenbaum

tapadipti:
Also, I didn't check which PR you had added your question in. So, I guess @daavoo's reply was enough. Anyway, I hope the above explanation doesn't hurt either 😄

PR #238, opened 2023-08-21
Separate eval into 2 commits (dvclive->custom)

dberenbaum:
Separates `eval` stage into 2 commits:

1. `dvclive-eval`: uses default dvclive configuration.
2. `custom-eval`: adds the custom plots config and sets `dvcyaml=False` in dvclive.

I'm working on accompanying docs changes to address https://github.com/iterative/dvc.org/issues/4769, with the idea that we introduce dvclive first and mention the custom plots config as an advanced next step.


example-get-started/generate.sh
```diff
@@ -301,16 +301,26 @@ EOF
   if [ $OPT_DVC_TRACKED_METRICS == "true" ]; then
     dvc stage add -n evaluate \
       -d src/evaluate.py -d model.pkl -d data/features \
-      -m eval/live/metrics.json -o eval/live/plots \
+      -o eval/live/metrics.json -o eval/live/plots \
       python src/evaluate.py model.pkl data/features
   else
     dvc stage add -n evaluate \
       -d src/evaluate.py -d model.pkl -d data/features \
-      -M eval/live/metrics.json -O eval/live/plots \
```
shcheklein, line 309: do we need to add `-O eval/live/metrics.json` here?

example-get-started/generate.sh
```diff
@@ -301,16 +301,26 @@ EOF
   if [ $OPT_DVC_TRACKED_METRICS == "true" ]; then
     dvc stage add -n evaluate \
       -d src/evaluate.py -d model.pkl -d data/features \
-      -m eval/live/metrics.json -o eval/live/plots \
+      -o eval/live/metrics.json -o eval/live/plots \
```
shcheklein, line 304: it should be `-O` by default I think (`cache: false`) - just to double check

example-get-started/generate.sh
shcheklein: need also to double check the semantics - `-o` vs `-O`

dberenbaum: No, and adding it would be inconsistent with the approach in https://github.com/iterative/dvclive/issues/456 and https://github.com/iterative/dvclive/pull/679. I don't have a strong opinion that we need to drop these outputs, but whatever convention we suggest, let's try to be consistent with it.

example-get-started/generate.sh
dberenbaum: Right, here because `$OPT_DVC_TRACKED_METRICS == "true"`, we use `-o`. Otherwise, we can use `-O` or drop them completely (see above).

example-get-started/generate.sh
shcheklein: kk, I'm fine with that. That was to double check primarily.

dberenbaum:
@shcheklein @daavoo Could you approve if you don't have any concerns? Would like to move forward with the docs changes, but they depend on these.

shcheklein (approved):

dberenbaum:
@shcheklein Redeployed it. Let me know if you think it's worth redeploying the fixtures and if you need help there.

shcheklein:
Tbh, I don't think so. It can wait until we change something more substantial.

dberenbaum:
Yup, sounds good to me

PR #239, opened 2023-08-22
add gha to test example-get-started

dberenbaum:
Step 1 in trying to setup some automation of this repo. This PR adds only an action to test that `example-get-started/generate.sh` runs successfully when changes to `example-get-started` are pushed. Next would be to make another action to do all the deployment steps when a PR is merged, and then we could add similar actions for `example-get-started-experiments`.

### DIFF
--- a/.github/workflows/example-get-started-test.yaml
+++ b/.github/workflows/example-get-started-test.yaml
@@ -1,0 +1,24 @@
+name: example-get-started test
+on: 
+  push:
+    paths:
+      - example-get-started/**
+  workflow_dispatch:
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.11'
+        cache: 'pip'
+    - uses: aws-actions/configure-aws-credentials@v2
+      with:
+        aws-region: us-east-2
+        role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+        role-duration-seconds: 43200
+    - name: Generate repo
+      run: ./generate.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/example-get-started-test.yaml
```diff
@@ -0,0 +1,26 @@
...
+      uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.11'
+        cache: 'pip'
+    - uses: aws-actions/configure-aws-credentials@v2
+      with:
+        aws-region: us-east-2
+        role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
```
dberenbaum, line 21: Added to repo secrets

daavoo (approved):
Let's also do example-get-started-experiments (needs CML runner)

.github/workflows/example-get-started-test.yaml
daavoo: Is it the sandbox I assume?

daavoo: Wouldn't you need to add this repo in https://github.com/iterative/itops/blob/dd03f7752ee61f09487902ce1e4e1d1074393748/terraform/iterative/iam_sandbox.tf#L38 ?

dberenbaum: Yup, thanks. I was sure I missed something (probably more). That's why I'm starting small.

.github/workflows/example-get-started-test.yaml
```diff
@@ -0,0 +1,26 @@
+name: example-get-started test
+on: 
+  push:
+    paths:
+      - example-get-started/**
+  workflow_dispatch:
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
daavoo, line 9:
```suggestion
    runs-on: ubuntu-latest
    environment: aws
```

daavoo: I think is needed for iterative/itops#2308 to work

PR #243, opened 2023-08-28
initial gha to test example-get-started-experiments

dberenbaum:
Similar to the recent action added to generate `example-get-started`, this does the same for `example-get-started-experiments`. It also uses the cml runner to launch a self-hosted instance for training.

This requires a `PERSONAL_ACCESS_TOKEN` be added to the repo. @daavoo @shcheklein How did we set up the token for `example-get-started-experiments`? Can/should we use the same here?

### DIFF
--- a/.github/workflows/example-get-started-experiments-test.yaml
+++ b/.github/workflows/example-get-started-experiments-test.yaml
@@ -25,12 +25,12 @@ jobs:
         run: |
           cml runner launch --single \
             --labels=cml \
-            --cloud=${{ inputs.cloud || 'aws' }} \
-            --cloud-region=${{ inputs.region || 'us-east' }} \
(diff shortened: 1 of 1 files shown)


daavoo:
> How did we set up the token for example-get-started-experiments? Can/should we use the same here?

I think it is a PAT from the @iterative-olivaw account. Access should be in 1password

dberenbaum:
Thanks @daavoo! Added the token to the repo secrets.

daavoo (approved):
Looks like we can't really try from the P.R so lets merge and followup if something fails

dberenbaum:
Yup, that's been my approach 🤷. Once we have an initial version merged, we can test follow-ups more easily at least.

PR #248, opened 2023-09-01
fix(example-get-started): cache images, fix reports action

shcheklein:
Fixes (bumped today into this):

Screenshot 2023-08-31 at 5 12 46 PM

Screenshot 2023-08-31 at 5 12 27 PM

It's better to cache images I think. It creates a stronger example on how things could be managed (I got a question today if images could be cached or not). It's better since we cover a bit more advanced scenario + probably ppl will use it this way.

### DIFF
--- a/example-get-started/code/src/evaluate.py
+++ b/example-get-started/code/src/evaluate.py
@@ -100,13 +100,14 @@ def main():
         test, _ = pickle.load(fd)
 
     # Evaluate train and test datasets.
-    live = Live(EVAL_PATH)
-    evaluate(model, train, "train", live, save_path=EVAL_PATH)
-    evaluate(model, test, "test", live, save_path=EVAL_PATH)
-    live.make_summary()
+    with Live(EVAL_PATH, cache_images=True, dvcyaml=False) as live:
+        evaluate(model, train, "train", live, save_path=EVAL_PATH)
+        evaluate(model, test, "test", live, save_path=EVAL_PATH)
 
-    # Dump feature importance plot.
-    save_importance_plot(live, model, feature_names)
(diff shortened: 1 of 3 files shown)


example-get-started/code/src/evaluate.py
```diff
@@ -100,13 +100,14 @@ def main():
...
-    evaluate(model, train, "train", live, save_path=EVAL_PATH)
-    evaluate(model, test, "test", live, save_path=EVAL_PATH)
-    live.make_summary()
+    with Live(EVAL_PATH, cache_images=True, dvcyaml=False) as live:
+        evaluate(model, train, "train", live, save_path=EVAL_PATH)
+        evaluate(model, test, "test", live, save_path=EVAL_PATH)
 
-    # Dump feature importance plot.
-    save_importance_plot(live, model, feature_names)
+        live.make_summary()
```
daavoo, line 107: if we do context manager then we don't need to explicitly call this
```suggestion
```

daavoo (approved):

example-get-started/code/src/evaluate.py
dberenbaum: I think it was intentional because we don't want to make dvc.yaml, report, etc. It is a bit advanced but we will either have to do it this way or enable some options like `with Live(dvcyaml=False)`.

daavoo: Then we need to add `report=None` on top of this because we would be now generating the report

dberenbaum: You mean if we use the context manager, we need `with Live(dvcyaml=False, report=None)`?

PR #251, opened 2023-09-08
fix(example-get-started): track the whole eval dir for simplicity

### LINKED ISSUE #249: `example-get-started`: cache images
> It's better to cache images I think. It creates a stronger example on how things could be managed (I got a question today if images could be cached or not). It's better since we cover a bit more advanced scenario + probably ppl will use it this way.

Originally from https://github.com/iterative/example-repos-dev/pull/248#issue-1876465827

shcheklein:
Fixes #249 

Simplify eval stage by tracking the whole eval directory with metrics, plots, etc.

Probably next step should be to get rid of the custom name, use log_artifact for models.

Tested it here https://github.com/shcheklein/example-get-started

### DIFF
--- a/example-get-started/code/.github/workflows/cml.yaml
+++ b/example-get-started/code/.github/workflows/cml.yaml
@@ -27,6 +27,7 @@ jobs:
             git fetch origin main:main
           fi
 
+          dvc pull eval
           dvc plots diff $PREVIOUS_REF workspace \
             --show-vega --targets ROC | json5 > vega.json
           vl2svg vega.json roc.svg
@@ -39,11 +40,9 @@ jobs:
             --show-vega --targets Confusion-Matrix | json5 > vega.json
           vl2svg vega.json confusion.svg
 
-          dvc pull eval/plots/images
           cp eval/plots/images/importance.png importance_workspace.png
 
           git checkout $PREVIOUS_REF -- dvc.lock
-          dvc pull eval/plots/images
           cp eval/plots/images/importance.png importance_previous.png
 
           dvc_report=$(dvc exp diff $PREVIOUS_REF --md)
--- a/example-get-started/code/README.md
+++ b/example-get-started/code/README.md
@@ -57,12 +57,14 @@ $ dvc pull
 
 ## Running in your environment
 
-Run [`dvc repro`](https://man.dvc.org/repro) to reproduce the
-[pipeline](https://dvc.org/doc/commands-reference/pipeline):
+Run [`dvc exp run`](https://man.dvc.org/exp/run) to reproduce the
+[pipeline](https://dvc.org/doc/user-guide/pipelines) and create a new
+[experiment](https://dvc.org/doc/user-guide/experiment-management).
 
 ```console
-$ dvc repro
-Data and pipelines are up to date.
+$ dvc exp run
+Ran experiment(s): rapid-cane
+Experiment results have been applied to your workspace.
 ```
 
 If you'd like to test commands like [`dvc push`](https://man.dvc.org/push),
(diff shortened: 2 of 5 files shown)


example-get-started/code/.github/workflows/cml.yaml
```diff
@@ -27,6 +27,7 @@ jobs:
             git fetch origin main:main
           fi
 
+          dvc pull eval
```
shcheklein, line 30: I'm not sure why we need this now (seems a regression, w/o it metrics are not shown in the `dvc exp diff` below) cc @dberenbaum . Also, I'm not sure I understand why this command is enough to get both revisions.

dberenbaum: Are you sure dvc-tracked metrics were ever pulled automatically? Up until now, we so often assumed metrics were git-tracked that I'm not sure we ever paid enough attention to this.

dberenbaum: I guess we need to fix this if we plan to start recommending dvc-tracked metrics

example-get-started/code/README.md
```diff
@@ -57,12 +57,14 @@ $ dvc pull
...
+Run [`dvc exp run`](https://man.dvc.org/exp/run) to reproduce the
+[pipeline](https://dvc.org/doc/user-guide/pipelines) and create a new
+[experiment](https://dvc.org/doc/user-guide/experiment-management).
 
 ```console
-$ dvc repro
-Data and pipelines are up to date.
+$ dvc exp run
+Ran experiment(s): rapid-cane
+Experiment results have been applied to your workspace.
```
dberenbaum, line 67: No strong opinion on which we recommend here, but we still use `dvc repro` through the docs.

example-get-started/code/src/evaluate.py
```diff
@@ -100,7 +100,7 @@ def main():
         test, _ = pickle.load(fd)
 
     # Evaluate train and test datasets.
-    with Live(EVAL_PATH, cache_images=True, dvcyaml=False) as live:
+    with Live(EVAL_PATH, dvcyaml=False, report=None) as live:
```
dberenbaum, line 103: Very minor, but we could probably drop `report=None` since it at worst will be an extra file and `report=None` will be the default in dvclive 3.0.

example-get-started/code/.github/workflows/cml.yaml
dberenbaum: Let's open an issue after this is deployed since it should be easy to reproduce from the deployed repo

dberenbaum (approved):
Thanks @shcheklein!

example-get-started/code/.github/workflows/cml.yaml
shcheklein: no, I'm not sure at all. Agreed that we need to fix this. Becomes a priority I guess. Since we always though had to cache them I would not be surprised if at some point we were fetching them.

dberenbaum: One more thought: haven't had a chance to test, but it might also work with stage-level metrics but not top-level metrics.

example-get-started/code/README.md
shcheklein: yes, I plan to change docs also when this is deployed - wdyt? (dvc repro becomes more of a low level command)

example-get-started/code/src/evaluate.py
shcheklein: yes, I'm following that ticket . I can change it when it's merged

example-get-started/code/README.md
dberenbaum: I go back and forth 😄 . Now I'm slightly in favor of keeping `repro` in this trail. Since we add one stage at a time, `dvc exp run` is a bit awkward to run after each stage is added. It's also unnecessarily heavy and the name doesn't fit the initial workflow we introduce very well. 

Most importantly, I don't think it is that impactful either way, so not something worth blocking.

shcheklein: okay, yes. I guess you are right. It can be ultimately a mix `dvc repro` to do iterations on the pipeline, but `dvc exp run` when everything is ready for that, wdyt?

dberenbaum: I'm not sure there's that much benefit to introduce a new command there, but it might make sense there, or as a "next step." Again, not a blocker for me, and probably better to discuss in a docs PR anyway where we can see how it actually looks.

example-get-started/code/.github/workflows/cml.yaml
dberenbaum: @skshetry WDYT about adding this to https://github.com/iterative/dvc/issues/9722?

shcheklein:
@dberenbaum another downside of tracking metrics is that we can't show a badge now:

Screenshot 2023-09-09 at 3 48 35 PM

not a deal breaker, but a bit sad. I would say metrics file is really small, it would be nice to have it in the Git history

PR #257, opened 2023-09-25
Add parent-sha input

daavoo:
Changes for https://github.com/iterative/studio/pull/7773 + some fixes


dberenbaum (approved):
Seems unfortunate that GH requires this to trigger a workflow for a non-head ref, but LGTM

PR #258, opened 2023-09-25
GitHub Actions: Deploy Repos

dberenbaum:
Debugs #256 to make it work and adds a similar workflow for example-get-started-experiments

### DIFF
--- a/.github/workflows/example-get-started-test.yaml
+++ b/.github/workflows/example-get-started-test.yaml
@@ -4,6 +4,8 @@ on:
     paths:
       - example-get-started/**
   workflow_dispatch:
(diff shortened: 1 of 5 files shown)


.github/workflows/example-get-started-test.yaml
```diff
@@ -3,6 +3,9 @@ on:
   push:
     paths:
       - example-get-started/**
+    branches:
+      - '**'        # matches every branch
+      - '!master'   # excludes master
```
dberenbaum, line 8: Avoid duplication with deploy workflow

shcheklein (approved):

PR #270, opened 2023-10-12
Sagemaker deployment - same role to assume and for deployment

tibor-mach:
If I understand the logic correctly, we want to use the AWS_SAGEMAKER_ROLE in both the credentials action and in the deploy_model script.

In fact at this point both of these `vars` contain the same role I think (though I don't have the rights to see secrets and variables in this repo).

But this way it is a bit cleaner, especially if someone wants to replicate the workflow.

### DIFF
--- a/example-get-started-experiments/code/.github/workflows/deploy-model.yml
+++ b/example-get-started-experiments/code/.github/workflows/deploy-model.yml
@@ -25,7 +25,7 @@ jobs:
         fetch-depth: 0
     - name: "Parse GTO tag"
       id: gto
-      uses: iterative/gto-action@14723404a00bb0c1e759c02ffcd24279df5815c2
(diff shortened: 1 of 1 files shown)


dberenbaum:
@0x2b3bfa0 Is it intentional to have 2 different roles here? If not, is there a preference for one or the other?

0x2b3bfa0:
@dberenbaum, if `vars.AWS_SANDBOX_ROLE` works as expected, I would not use `vars.AWS_SAGEMAKER_ROLE` at all; I believe it's the same. 🙈

dberenbaum:
@tibor-mach Let's close/update this one so that we use `vars.AWS_SANDBOX_ROLE` everywhere instead.

tibor-mach:
Closing, covered by #271

PR #271, opened 2023-10-19
Model registry steps to the getting started

tibor-mach:


dberenbaum:
@tibor-mach I haven't had a chance to review this yet. Let me know if you need it merged to test anything out. Otherwise, maybe let's leave it in draft mode as we work on the docs since those should drive the reqs here.

tibor-mach:
@dberenbaum Sure, the main difference here now is that I added a new GH workflow which doesn't require you to set up AWS or any other service so it is easier to try it out.

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
```diff
@@ -0,0 +1,67 @@
...
+  # This job parses the git tag with the GTO GitHub Action to identify model registry actions
+  parse:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: "Parse GTO tag"
+      id: gto
+      uses: iterative/gto-action@v2
```
dberenbaum, line 31: Can we update the other workflow to match any changes we make here?

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
```diff
@@ -0,0 +1,67 @@
...
+      with:
+        python-version: '3.8'
+        cache: 'pip'
+        cache-dependency-path: requirements.txt
+
+    - run: pip install -r requirements.txt
+
+    # this step adds our project's DVC remote to the workflow environment
+    - name: Add DVC Remote
+      run:  dvc remote add -d --local storage s3://dvc-public/remote/get-started-pools
```
dberenbaum, line 60: I don't think we need this step

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
```diff
@@ -0,0 +1,67 @@
...
+  test-model-artifact:
+    needs: parse
+    # using the outputs from the "parse" job, we run this job only for actions
+    # in the model registry and only when the model was assigned to a stage called "test"
+    if: "${{ needs.parse.outputs.event == 'assignment' }} && ${{ needs.parse.outputs.stage == 'test' }}"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
```
dberenbaum, line 47: If we use `dvc artifacts get`, we should be able to drop this step.

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
dberenbaum, line 60: If we use `dvc artifacts get`, we should be able to drop all these steps as long as dvc is installed (which we should add as a step).

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    # this step adds our project's DVC remote to the workflow environment
+    - name: Add DVC Remote
+      run:  dvc remote add -d --local storage s3://dvc-public/remote/get-started-pools
+    
+    # this step uses DVC to download the model from our remote repository and deploys the model
+    # Model deployment is mocked here as it is specific to each deployment environment
+    - name: Get Model and Check Size
+      run: |
+        dvc get model/model.pth # this will download the right model artifact from the DVC remote
```
dberenbaum, line 66: Let's use `dvc artifacts get` here (we can get all the args from the parse step).

tibor-mach: Sure, we can use it here, but we won't be able to use it with the sagemaker deployment - that one needs an S3 URL (that is specific to Sagemaker). I would suggest actually adding the --show-url functionality to dvc artifacts.

dberenbaum: That's a good idea. Even without that, though, I think we can get the url inside the sagemaker python script with only the gto info like:

```python
repo = 'https://github.com/iterative/example-get-started-experiments.git'
artifact = dvc.api.artifacts_show(
    name,
    version=version,
    stage=stage,
    repo=repo,
)
model_data = 
    artifact['path'],
    rev=artifact['rev'],
    repo=repo,
)
```

`dvc.api.artifacts_show` wasn't available when this was introduced, but now it should help here. Then we can drop this:

 https://github.com/iterative/example-repos-dev/blob/2f7a36d237e88b39083edcfc7e06f2c4fab5f3c9/example-get-started-experiments/code/.github/workflows/deploy-model.yml#L63

dberenbaum: @efiop The github action also adds a local s3 remote to override the http remote:

https://github.com/iterative/example-repos-dev/blob/2f7a36d237e88b39083edcfc7e06f2c4fab5f3c9/example-get-started-experiments/code/.github/workflows/deploy-model.yml#L60

Can we add the `config` option to `get_url` so we can add this info in the python call instead?

dberenbaum: By the way, no strong opinion that we do it this way for the sagemaker example. We can keep as is, use `dvc artifacts get --show-url`, or use the DVC API like I show above. Just thinking through the possible options.

tibor-mach: Personally, I would keep it like it is for now and replace it with `dvc artifacts get` later if we add the `--show-url` option. The sagemaker pipeline is not an integral part of the getting-stared guide, it is just another example we link to.

But the dvc.api approach is not visible form the workflow itself so it would require extra explanation.

Either way, I'd say it is not all that important what we do with the sagemaker workflow right now.

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
```diff
@@ -0,0 +1,54 @@
...
+  deploy-model:
+    needs: parse
+    # using the outputs from the "parse" job, we run this job only for actions
+    # in the model registry and only when the model was assigned to a stage called "prod"
+    if: "${{ needs.parse.outputs.event == 'assignment' }} && ${{ needs.parse.outputs.stage == 'prod' }}"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
```
tibor-mach, line 47:
```suggestion
```

dberenbaum:
@tibor-mach Should we move it out of draft status?

dberenbaum (commented):
@tibor-mach I guess it's waiting for you to decide whether to keep the checkout action or the github url variable, right?

tibor-mach:
> @tibor-mach I guess it's waiting for you to decide whether to keep the checkout action or the github url variable, right?

Yes and I also want to check whether we need the GH checkout action....which I will do now :-)

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
```diff
@@ -0,0 +1,51 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Manual version name'
+        required: true
+        type: string
+  
+permissions:
+  contents: write
+  id-token: write
```
dberenbaum, line 18: I think we could drop these AFAIK. Looking [here](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), write permissions to `contents` are granted by default (not sure if we need it anyway here), and `id-token` is only needed when using AWS roles to authorize via OIDC.

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
```diff
@@ -0,0 +1,51 @@
...
+  push:
+    tags:
+      - "*"
+  # we can also run the workflow manually...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Manual version name'
+        required: true
+        type: string
```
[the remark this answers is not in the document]
tibor-mach: @dberenbaum I think we can remove it completely from both workflows. I think that @daavoo intended to have a manual trigger but it was probably never fully implemented...I just copied it without really studying it too much.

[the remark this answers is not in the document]
dberenbaum: I don't mind the manual trigger. I'm just confused by the `version` input.

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
tibor-mach: Agreed. Again, this was a mindless copypasta...These should probably stay in the Sagemaker workflow, or at least the id-token. I'll try and see what happens if I remove one or both.

dberenbaum: Yeah, I'm fine to drop them here and leave in Sagemaker if you don't want to keep iterating on every combination here.

tibor-mach: Deleting them from template, keeping in sagemaker

dberenbaum (approved):

example-get-started-experiments/code/.github/workflows/deploy-model-template.yml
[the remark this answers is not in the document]
tibor-mach: The manual trigger does not really do anything anyway since GTO won't parse any tags and so the second job will always be skipped... I think it makes more sense to just remove it entirely.

PR #273, opened 2023-10-22
feat(example-get-started): add an option for GDrive remote

shcheklein:
Adds GDrive remote option to test Studio. I've updated the fixture here: https://github.com/iterative/example-get-started-gdrive-private-fixture/


dberenbaum:
Are we okay having the app credentials in the repo? I guess it means anyone could technically use this remote with these credentials?

shcheklein:
> Are we okay having the app credentials in the repo? I guess it means anyone could technically use this remote with these credentials?

No, those are app id secret and id only. One needs to auth with that app + have access to specific resources.

We have the same default app creds for prod DVC app https://github.com/iterative/dvc-gdrive/blob/main/dvc_gdrive/__init__.py#L23

dberenbaum (approved):

PR #274, opened 2023-11-01
Change first version from 'v0.1.0' to 'v1.0.0' to be consistent with …

jellebouwman:
…Studio suggestions

Let me know if I can help regenerating the projects 👍


dberenbaum:
Thanks @jellebouwman! I should have mentioned that the whole model registry workflow is getting updated in https://github.com/iterative/example-repos-dev/pull/271, and it looks like @tibor-mach already updated it there to `v1.0.0`, so I think we can close this in favor of that PR.

PR #275, opened 2023-11-10
chore(mlem): drop unmaintained fixture

shcheklein:
We don't use MLEM anymore. We don't need this fixture. Studio should migrate completely to regular DVC+GTO repos generated via other repos (e.g. `example-get-started`) has now a sophisticated way to generate nested repos with different models, etc.


shcheklein:
@dberenbaum ping

dberenbaum (approved):

PR #277, opened 2023-11-16
updates vars and secrets in ci

dberenbaum:
Attempts to fix:
- https://github.com/iterative/example-get-started-experiments/actions/runs/6896818532
- https://github.com/iterative/example-get-started-experiments/actions/runs/6896818531

### DIFF
--- a/example-get-started-experiments/code/.github/workflows/deploy-model-sagemaker.yml
+++ b/example-get-started-experiments/code/.github/workflows/deploy-model-sagemaker.yml
@@ -26,6 +26,7 @@ jobs:
   deploy-model:
     needs: parse
     if: "${{ needs.parse.outputs.event == 'assignment' }}"
+    environment: cloud
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
--- a/example-get-started-experiments/code/.github/workflows/dvc-studio.yml
+++ b/example-get-started-experiments/code/.github/workflows/dvc-studio.yml
@@ -109,7 +109,7 @@ jobs:
       - name: Train
         env:
           REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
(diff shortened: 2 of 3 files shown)


example-get-started-experiments/code/.github/workflows/dvc-studio.yml
```diff
@@ -109,7 +109,7 @@ jobs:
       - name: Train
         env:
           REPO_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
-          STUDIO_TOKEN: ${{ secrets.STUDIO_TOKEN }}
+          DVC_STUDIO_TOKEN: ${{ secrets.DVC_STUDIO_TOKEN }}
```
dberenbaum, line 112: We were using a mix of `STUDIO_TOKEN` and `DVC_STUDIO_TOKEN` before. Now it should consistently use `DVC_STUDIO_TOKEN` everywhere.

@tibor-mach Minor, but maybe we should use the env var in the deployment example to match here and because it seems a bit more straightforward for the CI case (no real need to configure it globally for one-time use).

example-get-started-experiments/code/.github/workflows/deploy-model-sagemaker.yml
```diff
@@ -26,6 +26,7 @@ jobs:
   deploy-model:
     needs: parse
     if: "${{ needs.parse.outputs.event == 'assignment' }}"
+    environment: cloud
```
dberenbaum, line 29: In #271, we changed the role from `AWS_SAGEMAKER_ROLE` to `AWS_SANDBOX_ROLE` and dropped `environment: cloud`. I'm taking a guess here that the `AWS_SAGEMAKER_ROLE` exists and has the correct value and was failing because it's only scoped to the `cloud` environment, but I need @shcheklein to confirm.

shcheklein (approved):

PR #279, opened 2023-12-18
Deploy model from MR in GitLab

tibor-mach:
Adding a model registry CICD template for GitLab. Needed for [this docs PR](https://github.com/iterative/dvc.org/issues/4998) . For simplicity (to have nothing but that example in the `.gitlab-ci.yml` file I renamed the old `.gitlab-ci.yml`...I am wondering whether we need that old one at all. It seems to be used to run experiments via CML, so it is probably not useless, but I would rather put it somewhere else.

In GitLab there apparently cannot be multiple workflow files, so I am not sure how to do this best (to keep the MR example file as simple as possible).

Any ideas about that @mnrozhkov or @dberenbaum ?


mnrozhkov:
@tibor-mach thanks! If the old CI config is valid for deployment scenarios, we may keep it in a separate `dir` like `gitlab` |
WDYT?

tibor-mach:
@mnrozhkov It looks like it is not just valid, but even necessary for some of the tests to pass :-) But I like your idea, I would instead put the example template in the `gitlab` folder and give it a descriptive name. People are going to copy-paste it anyway.

Although it would also be good to have a test for its validity. I guess that can be a new issue (for the backlog most likely).

dberenbaum:
> It looks like it is not just valid, but even necessary for some of the tests to pass

Looks like it is because the new file was named `gitlab-ci.yml` instead of `.gitlab-ci.yml`.

example-get-started-experiments/code/gitlab-ci.yml
```diff
@@ -0,0 +1,39 @@
...
+  script: 
+  # Install GTO to parse model tags
+  - pip install gto
+  # This job parses the model tags to identify model registry actions
+  - echo "CI_COMMIT_TAG - ${CI_COMMIT_TAG}" 
+  - echo MODEL_NAME="$(gto check-ref ${CI_COMMIT_TAG} --name)" >> parse.env
+  - echo MODEL_VERSION="$(gto check-ref ${CI_COMMIT_TAG} --version)" >> parse.env
+  - echo MODEL_EVENT="$(gto check-ref ${CI_COMMIT_TAG} --event)" >> parse.env
+  - echo MODEL_STAGE="$(gto check-ref ${CI_COMMIT_TAG} --stage)" >> parse.env
+  # Print variables saved to parse.env
```
dberenbaum, line 20: Just curious: is this necessary or something nice to have?

tibor-mach:
> Looks like it is because the new file was named gitlab-ci.yml instead of .gitlab-ci.yml.
ah, my bad

dberenbaum:
Other than that, LGTM to me, thanks @tibor-mach! I'm fine with the idea to move the old workflow to another dir for now.

dberenbaum:
> I guess it makes sense to do it this way and not the other way around

Not sure I follow. What are the options?

tibor-mach:
> > I guess it makes sense to do it this way and not the other way around
> 
> Not sure I follow. What are the options?

The other option is to create a directory called (e.g.) `gitlab-templates` and put the new workflow there. It would actually make sense to put all gitlab files there...kind of like we do with `.github/workflows/`, except that GitLab does not support that structure (it has some `include` keywords which kind of replicate this but it makes the syntax somewhat more complicated).

dberenbaum:
Okay, so it's about which file to keep in `.gitlab-ci.yml`, right? Yes, I agree the new one is the better "default" since I don't think we directly reference the old one anywhere in the docs, and the new one is needed for the MR get started workflow.

dberenbaum (approved):

PR #280, opened 2024-01-06
experiments: make notebook more like the pipeline

dberenbaum:
This PR makes the notebook and pipeline code more similar by modifying the notebook code to use the same paths and otherwise reusing the code from the pipeline. The changes from the notebook to the pipeline were probably more realistic than making them both so similar, but it makes the repo harder to understand and maintain.

### DIFF
--- a/example-get-started-experiments/code/notebooks/TrainSegModel.ipynb
+++ b/example-get-started-experiments/code/notebooks/TrainSegModel.ipynb
@@ -199,8 +199,9 @@
     "train_arch = 'shufflenet_v2_x2_0'\n",
     "\n",
     "for base_lr in [0.001, 0.005, 0.01]:\n",
-    "    # initialize dvclive, optionally provide output path, and save results as a dvc experiment\n",
-    "    with Live(\"results/train\", save_dvc_exp=False, report=\"notebook\") as live:\n",
+    "    # initialize dvclive, optionally provide output path, and show report in notebook\n",
+    "    # don't save dvc experiment until post-training metrics below\n",
+    "    with Live(\"results/train\", report=\"notebook\", save_dvc_exp=False) as live:\n",
     "        # log a parameter\n",
     "        live.log_param(\"train_arch\", train_arch)\n",
     "        fine_tune_args = {\n",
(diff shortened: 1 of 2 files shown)


example-get-started-experiments/code/notebooks/TrainSegModel.ipynb
```diff
@@ -182,44 +166,41 @@
...
+   "metadata": {},
    "outputs": [],
    "source": [
     "train_arch = 'shufflenet_v2_x2_0'\n",
-    "models_dir = ROOT / \"models\"\n",
-    "models_dir.mkdir(exist_ok=True)\n",
-    "results_dir = ROOT / \"results\" / \"train\"\n",
     "\n",
     "for base_lr in [0.001, 0.005, 0.01]:\n",
     "    # initialize dvclive, optionally provide output path, and save results as a dvc experiment\n",
```
shcheklein, line 202: update the comment?

shcheklein:
so, what is the current behavior? 

- does it still log artifact into the nested dvc.yaml or does it mean that it won't be saving anything in the notebook now?
- no experiments in notebook?

example-get-started-experiments/code/notebooks/TrainSegModel.ipynb
```diff
@@ -237,19 +218,22 @@
...
     "            type=\"model\",\n",
     "            name=\"pool-segmentation\",\n",
     "            desc=\"This is a Computer Vision (CV) model that's segmenting out swimming pools from satellite images.\",\n",
     "            labels=[\"cv\", \"segmentation\", \"satellite-images\", \"unet\"],\n",
-    "        )\n"
+    "        )\n",
+    "\n",
+    "    # add additional post-training summary metrics\n",
+    "    with Live(\"results/evaluate\") as live:\n",
+    "        live.summary[\"dice_multi\"] = evaluate(learn)"
```
dberenbaum, line 236: Experiment still gets saved here. `dvc_exp_save=False` above is only needed to avoid having two experiments saved for each run since there are two different live instances for train and eval.

example-get-started-experiments/code/notebooks/TrainSegModel.ipynb
```diff
@@ -41,7 +22,10 @@
     "from fastai.vision.all import (Resize, SegmentationDataLoaders,\n",
     "                               imagenet_stats, models, unet_learner)\n",
     "from ruamel.yaml import YAML\n",
-    "from PIL import Image"
+    "from PIL import Image\n",
+    "\n",
+    "os.chdir(\"..\")\n",
```
dberenbaum, line 27: By changing to the root dir at the start, we end up with similar behavior to the pipeline and the notebook writes to `dvc.yaml` there.

dberenbaum:
Studio project to see how it looks: https://studio.iterative.ai/user/dberenbaum/projects/exp-notebook-0nufj91gh8

shcheklein (approved):

PR #281, opened 2024-01-31
drop GIT_COMMITTER_DATE

dberenbaum:
Fixes https://github.com/iterative/studio/issues/8951.

It looks like this must be a bug in dulwich or scmrepo. When `GIT_COMMITTER_DATE` is set, it causes `GIT_COMMITTER_NAME` to be ignored. This is not the case for CLI Git, and the dates and formats appear valid. I don't see a need for this var anywhere (`GIT_AUTHOR_DATE` seems to be enough for the commits), so it looks easiest to drop this.


shcheklein (approved):

pmrowla (approved):

dberenbaum:
For reference, looks like this is a known issue: https://github.com/jelmer/dulwich/blob/98f303bdb24bd31ec25a543f7ab7af1d6704fa8f/dulwich/repo.py#L981

PR #283, opened 2024-05-04
feat(example-get-started): add GitLab CI to publish reports

shcheklein:
Adds GitLab CI pipeline to publish reports.

Example:

https://gitlab.com/iterative.ai/cse_public/example-text-classifier
https://gitlab.com/iterative.ai/cse_public/example-text-classifier/-/merge_requests/2


dberenbaum (approved):
Approving since I don't think there are any blockers, but I have a couple questions:
1. Should we add a mention to the readme?
2. Not sure I follow why some files get included in `code.zip` and others don't.

shcheklein:
> Should we add a mention to the readme?

I was thinking about this, but that means ideally we automate this as well. So, let's postpone this for now.

> Not sure I follow why some files get included in code.zip and others don't.

some legacy cleanup, no changes in terms of including older files.

Thanks @dberenbaum for the review.

